PR #2, opened 2015-12-22
WIP: Automate firstboot

mnowaksuse:
Future-wise we need a way how to test JeOS images automatically (i.e. without the need to always go thru the ncurses dialogs) in e.g. OpenQA.

This pull request is a way to do it in a hackish way. Though it does it's job.

### DIFF
--- a/contrib/patch-jeos-image-with-autofirstboot.sh
+++ b/contrib/patch-jeos-image-with-autofirstboot.sh
@@ -1,0 +1,45 @@
+#!/bin/sh
+
+clean(){
+  umount $MNT
+  qemu-nbd -d /dev/nbd${DEV}
+  rmmod nbd
+}
+
+#set -x
+set -e
+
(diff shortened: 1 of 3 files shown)


sysrich:
Sorry, jeos-firstboot is used by SUSE Linux Enterprise and a solution this hackish can never be accepted there - especially as it would enable users to bypass the legally important EULA

openQA can, and does, use the application like a user so a solution like this shouldn't be needed either.. Closing, but thanks for the effort!

lnussel:
I'm still interested in the background of this submission. You mentioned openQA but JeOS is already tested in openQA and the whole point of openQA is to go through all the steps a human has to do. So why did you feel the need for such a change?

mnowaksuse:
I didn't know of openQA capabilities, now, thanks to Richard, I know. Sorry for the noise.

PR #3, opened 2016-01-19
Added a README with a link to OBS

mvidner:
I heard "jeos-firstboot" mentioned in my team, so I searched, looked, saw no README...


mvidner:
@lnussel This is a trivial one, isn't it?

lnussel:
this is richard's playground, it's not the real upstream repo yet. It's not forgotten!

mvidner:
@sysrich is this the right repo?

PR #16, opened 2017-04-24
Add udev-settle to wait all udev rules to be run

gmoro:
This is required as Hyper-V and Xen have different console
setups that are required to finish before this service starts
Related to bsc#987589 and bsc#1031618


gmoro:
needs more test before merging, may not fully fix the issue.

PR #17, opened 2017-04-25
Fix for uninitialised console and yast bootloader

gmoro:


files/usr/lib/jeos-firstboot
```diff
@@ -30,6 +30,12 @@ EULA_FILE=/etc/YaST2/licenses/base/license.txt
 
 stty_size() {
     set -- `stty size`; LINES=$1; COLUMNS=$2
+    # stty size can return zero when not ready or
+    # its a serial console
+    if [ "${COLUMNS}" = "0" || "${LINES}" = "0" ]; then
```
lnussel: curly brackets are not need here. Use -o instead of ||

files/usr/lib/jeos-firstboot
```diff
@@ -220,6 +226,15 @@ if [ -n "$password" ]; then
     echo "root:$password" | run /usr/sbin/chpasswd
 fi
 
+EFIVARS_DIR="/sys/firmware/efi"
+# Look for EFI dir to see if the machine is booted in UEFI mode
+modprobe efivars
+if ! [ "$(ls -A $EFIVARS_DIR)" ]; then
```
lnussel: Raymund recommends to check for /sys/firmware/efi/systab, that's also what yast does.

files/usr/lib/jeos-firstboot
```diff
@@ -220,6 +226,15 @@ if [ -n "$password" ]; then
...
+    sed -i -e 's/LOADER_TYPE=.*/LOADER_TYPE=grub2/g' /etc/sysconfig/bootloader 
+else
+    echo "Grub is configured for EFI"
```
lnussel: useless message IMO, just omit it

files/usr/lib/jeos-firstboot
```diff
@@ -220,6 +226,15 @@ if [ -n "$password" ]; then
     echo "root:$password" | run /usr/sbin/chpasswd
 fi
 
+EFIVARS_DIR="/sys/firmware/efi"
+# Look for EFI dir to see if the machine is booted in UEFI mode
+modprobe efivars
+if ! [ "$(ls -A $EFIVARS_DIR)" ]; then
+    sed -i -e 's/LOADER_TYPE=.*/LOADER_TYPE=grub2/g' /etc/sysconfig/bootloader
```
lnussel: run sed ... to make dry mode work

files/usr/lib/systemd/system/jeos-firstboot.service
```diff
@@ -11,6 +11,7 @@ DefaultDependencies=no
 Conflicts=shutdown.target
 After=systemd-remount-fs.service
 Before=systemd-sysusers.service sysinit.target shutdown.target
+Wants=systemd-udev-settle.service
```
lnussel: before of cycles here. those deps have been tuned back and forth several times.

files/usr/lib/jeos-firstboot
gmoro: ok, fixed

files/usr/lib/jeos-firstboot
gmoro: yeah, I looked again in the YasT code, was going with a comment somewhere saying if the dir is not empty is enough to know, but better to keep in sync with YasT implementation that is well thought

files/usr/lib/jeos-firstboot
gmoro: fixed

files/usr/lib/jeos-firstboot
gmoro: indeed, did that for the test and forgot to take it out

files/usr/lib/systemd/system/jeos-firstboot.service
gmoro: added that a long time ago, still not sure if it's needed, the problem happened to be another thing, will remove for now

gmoro:
second round :)

PR #27, opened 2018-05-14
Fix timezone configuration

Vogtinator:
systemd-firstboot only applies changes if the options were previously unset.

References: bsc#1092132


Vogtinator:
Superseded by #28.

PR #28, opened 2018-05-16
Run sevice before systemd-tmpfiles, bsc#1092132

gmoro:
needed to avoid automatic creation of config files
under /etc during initial boot


Vogtinator:
No cycle, no other issues?

I wonder because systemd-tmpfiles might be needed for some of the tools used by jeos-firstboot to run correctly, like snapper.

PR #31, opened 2018-06-14
Refine locale setup again

Vogtinator:
Use live-langset-data.

This required a rewiring of the jeos-firstboot.service so I also changed the way it's enabled to fix a different bug.

### DIFF
--- a/files/usr/lib/systemd/system/jeos-firstboot.service
+++ b/files/usr/lib/systemd/system/jeos-firstboot.service
@@ -7,20 +7,23 @@
 
 [Unit]
 Description=SUSE JeOS First Boot Wizard
-DefaultDependencies=no
-Conflicts=shutdown.target
-After=systemd-remount-fs.service
-Before=systemd-tmpfiles-setup.service sysinit.target shutdown.target
-ConditionPathIsReadWrite=/etc
-RequiresMountsFor=/.snapshots
-# Use the same file as yast
+
+# Same as YaST2-Firstboot.service here
+After=apparmor.service local-fs.target plymouth-start.service YaST2-Second-Stage.service
+Conflicts=plymouth-start.service
+Before=getty@tty1.service serial-getty@ttyS0.service serial-getty@ttyS1.service serial-getty@ttyS2.service
+Before=display-manager.service
 ConditionPathExists=/var/lib/YaST2/reconfig_system
 OnFailure=poweroff.target
 
+# jeos-firstboot starts before network and login though
+Before=network.service systemd-user-sessions.service
+
 [Service]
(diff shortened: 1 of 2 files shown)


files/usr/lib/systemd/system/jeos-firstboot.service
```diff
@@ -7,19 +7,25 @@
...
+Before=network.service systemd-user-sessions.service
 
 [Service]
-Environment=SYSTEMCTL_OPTIONS=--ignore-dependencies TERM=linux
 Type=oneshot
+Environment=SYSTEMCTL_OPTIONS=--ignore-dependencies TERM=linux
 RemainAfterExit=yes
+ExecStartPre=-/usr/bin/plymouth quit
 ExecStart=/usr/lib/jeos-firstboot
+ExecStartPost=/usr/bin/rm -f /var/lib/YaST2/reconfig_system
```
mnowaksuse: Are we going to have this file? (again)

files/usr/lib/systemd/system/jeos-firstboot.service
```diff
@@ -7,19 +7,25 @@
...
+
+# jeos-firstboot starts before network and login though
+Before=network.service systemd-user-sessions.service
 
 [Service]
-Environment=SYSTEMCTL_OPTIONS=--ignore-dependencies TERM=linux
 Type=oneshot
+Environment=SYSTEMCTL_OPTIONS=--ignore-dependencies TERM=linux
 RemainAfterExit=yes
+ExecStartPre=-/usr/bin/plymouth quit
```
mnowaksuse: Do we have `plymouth`?

files/usr/lib/systemd/system/jeos-firstboot.service
```diff
@@ -7,19 +7,25 @@
...
-ConditionPathIsReadWrite=/etc
-RequiresMountsFor=/.snapshots
-ConditionFirstBoot=yes
+
+# Same as YaST2-Firstboot.service here
+After=apparmor.service local-fs.target plymouth-start.service YaST2-Second-Stage.service
+Conflicts=plymouth-start.service
+Before=getty@tty1.service serial-getty@ttyS0.service serial-getty@ttyS1.service serial-getty@ttyS2.service
+Before=display-manager.service
+ConditionPathExists=/var/lib/YaST2/reconfig_system
```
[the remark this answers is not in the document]
Vogtinator: Yes, it's added by the image build.

files/usr/lib/systemd/system/jeos-firstboot.service
Vogtinator: Not on JeOS for SLE 12/15, but that might change in the future and/or be different for openSUSE - calling plymouth quit can't hurt here (the "-" means that it won't fail).

PR #32, opened 2018-07-20
kmscon integration

Vogtinator:
Known bugs:
- ~~Palette is ugly~~
- ~~Return value is wrong?~~

### DIFF
--- a/files/usr/lib/jeos-firstboot
+++ b/files/usr/lib/jeos-firstboot
@@ -191,7 +191,7 @@ if [[ "$(tty)" = /dev/tty[0-9]* ]]; then
 
         if kmscon_available; then
             ret_file="$(mktemp)"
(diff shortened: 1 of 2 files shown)


files/usr/lib/jeos-firstboot
```diff
@@ -131,62 +131,89 @@ findlocales()
...
-	for i in "${EULA_FILE%.txt}.$locale.txt" \
-	         "${EULA_FILE%.txt}.${locale%%.UTF-8}.txt" \
-	         "${EULA_FILE%.txt}.${language}.txt"; do
-	    if [ -e "$i" ]; then
-		EULA_FILE="$i"
-		break
-	    fi
-	done
+    declare -A start_kmscon
+    start_kmscon["cs"]=1
```
mnowaksuse: Why `cs`?

Vogtinator: license.cs.txt from openSUSE at least contains characters the console font does not contain.

PR #40, opened 2019-02-12
Fixes for wireless

gmoro:


Vogtinator:
I found some bugs - those were fixed in master already though. Can you rebase so I can test properly?

Vogtinator:
I wasn't able to test the wlan connection itself yet (wpa_supplicant missing), but I noticed that the behaviour of declining the EULA changed - can you just use "dialog" instead of "d" there?

Issue #4, opened 2016-08-11
allow custom wizard steps


lnussel:
jeos-firstboot should be able to execute files from a directory to allow adding custom wizard steps after the official ones. Useful for applicances that require customization for their software.

Vogtinator:
Implemented now.

PR #44, opened 2019-02-28
Execute external script(s) on firstboot

illuusio:
As in Issue #4 states: There should be way to execute external scripts. This is bit naive way to achieve that goal but should be enough for most cases.

### DIFF
--- a/files/usr/lib/jeos-firstboot
+++ b/files/usr/lib/jeos-firstboot
@@ -63,13 +63,13 @@ fi
 modules=()
 
 call_module_hook() {
-    hook="$1"
+    local hook="$1"
     for module in "${modules[@]}"; do
         hook_function="${module}_${hook}"
-        [ $(type -f "${hook_function}") == "function" ] || continue
-        "${hook_function}" && true # To not trigger errexit
+        [ $(type -f "${hook_function}") = "function" ] || continue
+        "${hook_function}" || true # To not trigger errexit
         ret=$?
-        [ ${ret} -eq 0 ] || return ${ret}
+        [ $ret -eq 0 ] || return $ret
     done
     return 0
 }
--- a/files/usr/lib/jeos-firstboot
+++ b/files/usr/lib/jeos-firstboot
@@ -489,8 +489,8 @@ if pushd "/usr/share/jeos-firstboot" &>/dev/null; then
     for module in *; do
         source "${module}" && modules+=("${module}")
     done
+    popd &>/dev/null
 fi
-popd &>/dev/null
 
 call_module_hook systemd_firstboot
 call_module_hook post
--- a/files/usr/lib/jeos-firstboot
+++ b/files/usr/lib/jeos-firstboot
@@ -67,7 +67,7 @@ call_module_hook() {
     for module in "${modules[@]}"; do
         hook_function="${module}_${hook}"
         [ $(type -f "${hook_function}") = "function" ] || continue
-        "${hook_function}" || true # To not trigger errexit
+        "${hook_function}" && true # To not trigger errexit
         ret=$?
         [ $ret -eq 0 ] || return $ret
     done
--- a/files/usr/lib/jeos-firstboot
+++ b/files/usr/lib/jeos-firstboot
@@ -361,6 +361,7 @@ for d in ${candidates[@]} ; do
         break # one is enough
 done
 
+config_wireless=false
 trap 'call_module_hook cleanup' EXIT
 call_module_hook systemd_firstboot
 
--- a/files/usr/lib/jeos-firstboot
+++ b/files/usr/lib/jeos-firstboot
@@ -388,9 +388,4 @@ if [ ! -e /.snapshots/2 ]; then
 fi
 run /usr/lib/snapper/plugins/grub --refresh
 
-if [ "$config_wireless" = "true" ]; then
-    run ifdown $wlan_device 2>/dev/null || true
-    run ifup $wlan_device || true
-fi
-
 # vim: sw=4
--- a/files/usr/lib/jeos-firstboot
+++ b/files/usr/lib/jeos-firstboot
@@ -352,111 +352,6 @@ for d in ${candidates[@]} ; do
         break # one is enough
 done
 
-config_wireless=false
-if is_raspberry && ls -d /sys/class/net/*/wireless &>/dev/null; then
-    if dialog --yesno $"Configure wireless network?" 0 0; then
-	config_wireless=true
-    fi
-fi
-if [ "$config_wireless" = "true" ]; then
-    # Wi-fi
-    #
-    get_wlan_devices()
-    {
-        list=()
-        local line
-        while read line; do
-                list+=("${line}" '') 
-        done < <(ls -d /sys/class/net/*/wireless | awk -F'/' '{ print $5 }')
-        [ -n "$list" ]
-    }
-
-    get_wlan_networks()
-    {
-        list=()
-        local line
-        while read line; do
-                list+=("${line}" '') 
-        done < <(ip link set $wlan_device up && iwlist $wlan_device scan|grep ESSID|cut -d':' -f2|cut -d'"' -f2|sort -u)
-        [ -n "${list[*]}" ]
-    }
-
-    wlan_error(){
-        d --title $"Error" --msgbox "$1" 0 0
-        dialog --yesno $"Do you want to retry?" 0 0
-    }
-    while true
-    do
-        if get_wlan_devices && d --menu  $"Select Wireless card to configure" 0 0 $dh_menu "${list[@]}"; then
-            wlan_device="$result"
-        else
-            if wlan_error $"Error listing wlan devices"; then
-                continue
-            else
-                break
-            fi
-        fi
-
-        if get_wlan_networks && d --menu  $"Select Wireless network to connect" 0 0 $dh_menu "${list[@]}"; then
-            wlan_network="$result"
-        else
-            if wlan_error $"Error listing wireless networks"; then
-                continue
-            else
-                break
-            fi
-        fi
-
-        if d --menu $"Select authentication mode" 0 0 $dh_menu $"WPA-PSK" '' $"WPA-EAP" '' $"Open" ''; then
-            wlan_auth_mode="$result"
-        fi
-        wlan_auth_mode_conf=
-        if [ "$wlan_auth_mode" = "WPA-EAP" ]; then
-            wlan_auth_mode_conf=eap
-            if d --inputbox $"Username" 0 0; then
-                wlan_username="$result"
-            fi
-        fi
-        if [ "$wlan_auth_mode" = "WPA-PSK" ]; then
-                wlan_auth_mode_conf=psk
-        fi
-        if [ "$wlan_auth_mode" = "Open" ]; then
-                wlan_auth_mode_conf=open
-        fi
-        if [ "$wlan_auth_mode" != "Open" ]; then	
-            wlan_password=
-            if d --insecure --passwordbox  $"Password" 0 0; then
-                wlan_password="$result"
-            fi
-        fi
-        break
-    done
- 
-    config_file=`mktemp -qt 'firstboot-XXXXXX'`
-
-    cat << EOF > $config_file
-BOOTPROTO='dhcp'
-STARTMODE='auto'
-WIRELESS_AP_SCANMODE='1'
-WIRELESS_AUTH_MODE='$wlan_auth_mode_conf'
-WIRELESS_ESSID='$wlan_network'
-WIRELESS_MODE='Managed'
-EOF
-
-    if [ $wlan_auth_mode = "WPA-PSK" ]; then
-        echo "WIRELESS_WPA_PSK='$wlan_password'" >> $config_file
-    fi
-
-    if [ $wlan_auth_mode = "WPA-EAP" ]; then
-        echo "WIRELESS_WPA_IDENTITY='$wlan_username'" >> $config_file
-        echo "WIRELESS_WPA_PASSWORD='$wlan_password'" >> $config_file
-        echo "WIRELESS_EAP_AUTH='mschapv2'" >> $config_file
-    fi
-
-    run mv -f $config_file /etc/sysconfig/network/ifcfg-$wlan_device
-	
-fi
-
 d --infobox $"Applying firstboot settings ..." 3 40 || true
 # FIXME: systemd-firstboot doesn't set password if shadow present
 if [ -n "$password" ]; then
@@ -479,11 +374,6 @@ if [ ! -e /.snapshots/2 ]; then
     run create_snapshot 2 "Initial Status" "yes" || true
 fi
 run /usr/lib/snapper/plugins/grub --refresh
-
-if [ "$config_wireless" = "true" ]; then
-    run ifdown $wlan_device 2>/dev/null || true
-    run ifup $wlan_device || true
-fi
 
 if pushd "/usr/share/jeos-firstboot" &>/dev/null; then
     for module in *; do
--- a/files/usr/lib/jeos-firstboot
+++ b/files/usr/lib/jeos-firstboot
@@ -361,9 +361,9 @@ for d in ${candidates[@]} ; do
         break # one is enough
 done
 
+trap 'call_module_hook cleanup' EXIT
 call_module_hook systemd_firstboot
 call_module_hook post
-call_module_hook cleanup
 
 d --infobox $"Applying firstboot settings ..." 3 40 || true
 # FIXME: systemd-firstboot doesn't set password if shadow present
--- a/files/usr/lib/jeos-firstboot
+++ b/files/usr/lib/jeos-firstboot
@@ -470,4 +470,13 @@ if [ "$config_wireless" = "true" ]; then
     run ifdown $wlan_device 2>/dev/null || true
     run ifup $wlan_device || true
 fi
+
+# Execute external script(s) on firstboot
+if [ -d "/usr/lib/jeos-firstboot.d" ]; then
+    for script in /usr/lib/jeos-firstboot.d/??-*; do
+        if [ -x "${script}" ]; then
+            run "${script}"
+        fi
+    done
+fi
 # vim: sw=4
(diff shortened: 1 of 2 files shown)


Vogtinator:
That's indeed a naive way, those scripts aren't run at all if firstboot is run in dry-run mode.

This addition would be a externally visible API, so IMO we should make it slightly more refined before making it available.

illuusio:
Yes but nothing is run in dry-run? So I though that would be as it should :dancer:.
Just curious what do you mean visible externally visible API? Do you meand SELinux?

Vogtinator:
> Yes but nothing is run in dry-run? So I though that would be as it should.

Not quite right - it still shows dialogs and computes the commands, even if they're not executed. That makes it possible to test it without making changes to the running system.

IMO something like this should be implemented by sourcing (instead of executing) files in the specified directories (I'd say `/etc/jeos-firstboot` in addition as well, for custom scripts not in official packages). That would need to be documented, along with all variables and functions the scripts can use (like `run` and `d`).

> Just curious what do you mean visible externally visible API? Do you meand SELinux?

If jeos-firstboot calls external scripts, that's an API. jeos-firstboot changes in the future shouldn't break those scripts in a backwards-incompatible way.

illuusio:
Ok this can be closed I suppose and rethinked?

Vogtinator:
We can leave this open and discuss here as well, what do you prefer?

illuusio:
Ok we can do that also. This is needed feature for me.

Vogtinator:
I created a wiki page to collect ideas: https://github.com/openSUSE/jeos-firstboot/wiki/Support-for-extensions-using-custom-scripts

illuusio:
I could be all wrong but should it be something like this (I'll draft this on wiki after this if needed):

- Source script with --dry  if needed (i.e . script --dry). Because they are sources they should allow using function d, run and others.
- If script returns 0 it's success and 1 if not. Should everything be stopped after error?
- There could be need for functions like (Names can be anything): before(), execute() and after(). These could be called from master script and not just trust it will be executed or is just KISS exceute script and be happy.
- Could they return new global variables?

Vogtinator:
Yeah, using hook functions is probably the best approach here.

I'm thinking about something like:

```
# Load all available modules
pushd /usr/share/jeos-firstboot
for module in *; do
    source $module && modules+=("$module")
done
popd

...
# Run all systemd_firstboot hooks
for module in "${module[@]}"; do
    [ $(type -f "${module}_systemd_firstboot") = "function" ] || continue
    "${module}_systemd_firstboot"
done
```

illuusio:
Ok now we are talking about serious BASH wizarding :+1:. How do to write that in Wiki?

Vogtinator:
I added some ideas to the page - if you have anything else, just add it.

illuusio:
I made changes to have more advanced version of script launching. It should more like what has risen in this discussion and wiki page.

files/usr/lib/jeos-firstboot
```diff
@@ -470,4 +470,30 @@ if [ "$config_wireless" = "true" ]; then
     run ifdown $wlan_device 2>/dev/null || true
     run ifup $wlan_device || true
 fi
+
+if [ -d "/usr/share/jeos-firstboot" ]; then
```
Vogtinator: This would be nicer:

```
modules=()

if pushd "/usr/share/jeos-firstboot" &>/dev/null; then
        for module in *; do
        source $module && modules+=("$module")
    done
popd &>/dev/null
```

Then the rest doesn't need to be inside the if. I'd prefer it if we had a generic `call_module_hook` function, like this:

```
call_module_hook() {
    hook="$1"
    for module in "${module[@]}"; do
        [ $(type -f "${module}_$hook") == "function" ] || continue
        "${module}_$hook" && true # To not trigger errexit
        [ $? -eq 0 ] || return $?
    done
    return 0
}
```

and could use this in the appropriate places.

illuusio: Should call_module_hook be in functions file?

illuusio: And is this little bit wrong as it should be hook1_function, hook1_post and hook1_cleanup and after that hook2_function, hook2_post and hook2_cleanup. As this is just cosmetic and actually doesn't make any difference.

Vogtinator: My idea about the `_post` and `_cleanup` was different - it's supposed to run in appropriate places in `jeos-firstboot`, not for each hook.

`jeos-firstboot` would contain

`trap EXIT "call_module_hook cleanup"` for example.

files/usr/lib/jeos-firstboot
```diff
@@ -471,29 +485,14 @@ if [ "$config_wireless" = "true" ]; then
...
-        fi
-
-        # Check if cleanup function is available
-        if [ $(type -f "${module}_cleanup") = "function" ]; then
-            "${module}_cleanup"
-        fi
+        source "${module}" && modules+=("${module}")
     done
 fi
+popd &>/dev/null
```
Vogtinator: Needs to be inside the if

files/usr/lib/jeos-firstboot
illuusio: Ok i've little bit updated script to fit this idea. I don't know it trap exit is needed as call_module_hook checks if there is cleanup or post function?

files/usr/lib/jeos-firstboot
illuusio: My bad! Was testing in other machine and didn't update in git.

files/usr/lib/jeos-firstboot
```diff
@@ -60,6 +60,20 @@ else
     }
 fi
 
+modules=()
+
+call_module_hook() {
+    hook="$1"
```
lnussel: local

files/usr/lib/jeos-firstboot
```diff
@@ -60,6 +60,20 @@ else
...
+    for module in "${modules[@]}"; do
+        hook_function="${module}_${hook}"
+        [ $(type -f "${hook_function}") == "function" ] || continue
```
lnussel: This is shell. use = instead of ==

files/usr/lib/jeos-firstboot
```diff
@@ -60,6 +60,20 @@ else
...
+        "${hook_function}" && true # To not trigger errexit
```
lnussel: ||

files/usr/lib/jeos-firstboot
```diff
@@ -60,6 +60,20 @@ else
...
+        ret=$?
+        [ ${ret} -eq 0 ] || return ${ret}
```
lnussel: useless { }

files/usr/lib/jeos-firstboot
```diff
@@ -470,4 +484,15 @@ if [ "$config_wireless" = "true" ]; then
     run ifdown $wlan_device 2>/dev/null || true
     run ifup $wlan_device || true
 fi
+
+if pushd "/usr/share/jeos-firstboot" &>/dev/null; then
+    for module in *; do
+        source "${module}" && modules+=("${module}")
```
lnussel: if . "$module"; then
 ...

illuusio:
Now should be reviews updated

files/usr/lib/jeos-firstboot
```diff
@@ -63,13 +63,13 @@ fi
...
 
 call_module_hook() {
-    hook="$1"
+    local hook="$1"
     for module in "${modules[@]}"; do
         hook_function="${module}_${hook}"
-        [ $(type -f "${hook_function}") == "function" ] || continue
-        "${hook_function}" && true # To not trigger errexit
+        [ $(type -f "${hook_function}") = "function" ] || continue
+        "${hook_function}" || true # To not trigger errexit
```
Vogtinator: That way `ret` is always 0.

Vogtinator:
Looking good so far (except the broken `|| true` now), so only actually defining and using hooks needs to be done.

I guess we should define the format arguments can be supplied to systemd firstboot as a global `systemd_firstboot_args` variable directly.

files/usr/lib/jeos-firstboot
lnussel: what is the function supposed to do in the error case anyways? you could of course do
   "$hook_function" || return $?

but then the script would die later. So probably printing some error and continuing is better

illuusio:
All global variables should be available because they are run under main script? Should I turn that WIFI stuff as a script for a test so this can be evaluated to work?

Vogtinator:
> Should I turn that WIFI stuff as a script for a test so this can be evaluated to work?

Yes, that would be a good example and testcase.

files/usr/lib/jeos-firstboot
illuusio: I'll vote for error printing and then in distant future when there is more time and knownledge script would handle errors somehow..

illuusio:
Ok I turned WIFI to script. I have to make RPM from this that I can test it on image.

files/usr/lib/jeos-firstboot
```diff
@@ -466,8 +375,16 @@ if [ ! -e /.snapshots/2 ]; then
 fi
 run /usr/lib/snapper/plugins/grub --refresh
 
-if [ "$config_wireless" = "true" ]; then
-    run ifdown $wlan_device 2>/dev/null || true
-    run ifup $wlan_device || true
+if pushd "/usr/share/jeos-firstboot" &>/dev/null; then
```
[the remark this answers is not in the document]
illuusio: You mean call_module_hook systemd_firstboot? Ain't post enough?

illuusio:
Now it does make snapper after scripts as it should be?

files/usr/lib/jeos-firstboot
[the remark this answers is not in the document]
Vogtinator: It should be moved to line 77 in the file.

files/usr/lib/jeos-firstboot
```diff
@@ -466,8 +375,16 @@ if [ ! -e /.snapshots/2 ]; then
...
+    for module in *; do
+        if . "${module}"; then
+            modules+=("${module}")
+        fi
+    done
+    popd &>/dev/null
 fi
+
+call_module_hook systemd_firstboot
```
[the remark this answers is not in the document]
Vogtinator: AFAICT this wasn't fixed.

illuusio:
I reverted stuff and done it (hopefully) as expected

files/usr/lib/jeos-firstboot
```diff
@@ -470,4 +392,5 @@ if [ "$config_wireless" = "true" ]; then
     run ifdown $wlan_device 2>/dev/null || true
```
Vogtinator: This needs to be part of the raspberrywifi's post hook.

files/usr/lib/jeos-firstboot
```diff
@@ -338,110 +361,9 @@ for d in ${candidates[@]} ; do
...
-        echo "WIRELESS_WPA_PASSWORD='$wlan_password'" >> $config_file
-        echo "WIRELESS_EAP_AUTH='mschapv2'" >> $config_file
-    fi
-
-    run mv -f $config_file /etc/sysconfig/network/ifcfg-$wlan_device
-	
-fi
+call_module_hook systemd_firstboot
+call_module_hook post
+call_module_hook cleanup
```
Vogtinator: cleanup needs to run on exit, use `trap` for that.

illuusio: Haven't used trap for while but now it should be updated

files/usr/lib/jeos-firstboot
```diff
@@ -361,9 +361,9 @@ for d in ${candidates[@]} ; do
         break # one is enough
 done
 
+trap 'call_module_hook cleanup' EXIT
 call_module_hook systemd_firstboot
 call_module_hook post
```
Vogtinator: This needs to be moved to the place where the code previously was.

illuusio: Actually not if we want to make it run always on exit even if something very bad happens. As script is only run only at the end.

Vogtinator: I'm referring to `call_module_hook post`, not the cleanup trap.

illuusio: Did I messed something up?

illuusio: Now it should be in correct place?

Vogtinator: Yup!

files/usr/share/jeos-firstboot/raspberrywifi
```diff
@@ -0,0 +1,141 @@
...
+
+raspberrywifi_systemd_firstboot()
+{
+    config_wireless=false
+    if is_raspberry && ls -d /sys/class/net/*/wireless &>/dev/null; then
+        if dialog --yesno $"Configure wireless network?" 0 0; then
+            config_wireless=true
+        fi
+    fi
+    if [ "$config_wireless" = "true" ]; then
```
Vogtinator: The whole "if" can be moved one layer up, like this:

```
if foo; then
   bar;
fi
```

to 

```
foo || return
bar
```

files/usr/share/jeos-firstboot/raspberrywifi
```diff
@@ -0,0 +1,141 @@
...
+
+raspberrywifi_wlan_error()
+{
+    d --title $"Error" --msgbox "$1" 0 0
+    dialog --yesno $"Do you want to retry?" 0 0
+}
+
+raspberrywifi_systemd_firstboot()
+{
+    config_wireless=false
```
Vogtinator: This is a global variable - please move the initialization it into global scope

files/usr/share/jeos-firstboot/raspberrywifi
```diff
@@ -0,0 +1,141 @@
...
+
+raspberrywifi_post()
+{
+    if [ "$config_wireless" = "true" ]; then
+        run ifdown $wlan_device 2>/dev/null || true
+        run ifup $wlan_device || true
+    fi
+}
+
+raspberrywifi_cleanup()
```
Vogtinator: Is a no-op, so can be removed

Vogtinator:
Can you also add some short comments on each hook implementation in the module, e.g.

```
# This is called by jeos-firstboot for user interaction and access to the global systemd_firstboot_args array
raspberrywifi_systemd_firstboot() {
...
}

File: files/usr/share/jeos-firstboot/raspberrywifi (hunk 14 above)
Comment (illuusio) [reply files/usr/share/jeos-firstboot/raspberrywifi]: I'm not sure that it's working as expected. Should there be some == or something to make sure it's working before ||

File: files/usr/share/jeos-firstboot/raspberrywifi [hunk 17]
```diff
@@ -0,0 +1,138 @@
...
+    if [ $wlan_auth_mode = "WPA-EAP" ]; then
+        echo "WIRELESS_WPA_IDENTITY='$wlan_username'" >> $config_file
+        echo "WIRELESS_WPA_PASSWORD='$wlan_password'" >> $config_file
+        echo "WIRELESS_EAP_AUTH='mschapv2'" >> $config_file
+    fi
+
+    run mv -f $config_file /etc/sysconfig/network/ifcfg-$wlan_device
+}
+
+# This is called by jeos-firstboot for user
```
Vogtinator: Nope

files/usr/share/jeos-firstboot/raspberrywifi
```diff
@@ -0,0 +1,138 @@
...
+# interaction and access to the global systemd_firstboot_args array
+raspberrywifi_systemd_firstboot()
+{
+    if is_raspberry && ls -d /sys/class/net/*/wireless &>/dev/null; then
+        if dialog --yesno $"Configure wireless network?" 0 0; then
+            config_wireless=true
+        fi
+    fi
+    $config_wireless || return
```
Vogtinator:
```suggestion
    [ "$config_wireless" = "true" ] || return
```

Vogtinator:
Almost there! Next week @gmoro is back and will have a look as well.

files/usr/share/jeos-firstboot/raspberrywifi
illuusio: Nope for comment?

Vogtinator: Yup - the comment for `_post` should be different than the one for `_systemd_firstboot`.

files/usr/share/jeos-firstboot/raspberrywifi
illuusio: Ah Github have too many options that I haven't used and this one is one of them. Corrected it by myself.

illuusio:
Hopefully. This has been serious lesson of Bash-fu. I though I knew something but learning is good thing.

files/usr/share/jeos-firstboot/raspberrywifi
Vogtinator: I don't like it really, it doesn't even allow multline changes yet (or I haven't found a way to make those).

files/usr/share/jeos-firstboot/raspberrywifi
illuusio: Ok now there is very cryptic comment

Vogtinator: This comment doesn't make much sense. What about

```
# This is called after the configuration steps finished successfully
```

illuusio:
I'll rebase if needed

Vogtinator:
I'd prefer if every module was split into a package with proper dependencies (e.g. WLAN configuration needs wifi tools) and only those modules actually used are installed. config.sh could remove modules it doesn't need.

Having a configuration file referenced via `EnvironmentFile=-/etc/jeos-firstboot.conf` would work as well, it would also allow setting some of the other exposed options.

aplanas:
> Having a configuration file referenced via `EnvironmentFile=-/etc/jeos-firstboot.conf` would work as well, it would also allow setting some of the other exposed options.

Even better. I agree.

illuusio:
So should that /etc/jeos-firstboot.conf contain enviromental variables which are read on systemd startup? Would it be just like JEOS_FIRSTBOOS_SCRIPTS="script1.sh script2.sh" or something fancier?

lnussel:
can't really spend time reviewing this atm but this clearly goes into a directory of systemd style handling of services.

PR #45, opened 2019-03-15
Support the non-standard and hopefully temporary location for the SLE EULAs

Vogtinator:
While openSUSE changed the main location for license files to the standard RPM
license directory /usr/share/licenses/$packagename/, SLE did it in an
incompatible way that requires indirection to get to it.

Works around jsc#SLE-3067 and fixes bsc#1127166


lnussel:
in theory you could read base_product_license_directory from /etc/YaST2/control.xml

PR #46, opened 2019-03-22
Move snapshot creation into a separate service

Vogtinator:
- This means the snapshot is created only after time is synced
- Fixes time of snapshot bsc#1129730

I'll remove the WIP flag once the reporter of the bug confirms that this approach fixes it.


Vogtinator:
Conflicts resolved, please review

PR #49, opened 2019-05-08
Allows the skip of timezone and password dialogs

aplanas:
Add two new variables to jeos-firstboot:

* `TIMEZONE_PRESET`
  Based on `LOCALE_PRESET` and `KEYTABLE_PRESET`, this new variable will affect the `systemd_firstboot_args` array if is set, and will skip the time-zone dialog.

* `PASSWORD_ALREADY_SET`
  If set to a non-zero value, will skip the root password dialog. Is expected that the user is setting the root password via other means, like `config.sh` or the Kiwi `users` XML tag.

Also fix the call of btrfs on non-btrfs file systems.

### DIFF
--- a/files/usr/lib/jeos-firstboot
+++ b/files/usr/lib/jeos-firstboot
@@ -296,14 +296,13 @@ if [ -z "$JEOS_TIMEZONE" ]; then
 		/usr/share/zoneinfo/zone.tab \
 	    && d --default-item "$default" --menu $"Select Time Zone" 0 0 $dh_menu "${list[@]}"; then
 	if [ -n "$result" ]; then
-	    systemd_firstboot_args+=("--timezone=$result")
+	    JEOS_TIMEZONE="$result"
 	fi
     else
 	d --msgbox $"error setting timezone" 0 0
     fi
-else
-    systemd_firstboot_args+=("--timezone=$JEOS_TIMEZONE")
 fi
+systemd_firstboot_args+=("--timezone=$JEOS_TIMEZONE")
 
 # systemd-firstboot does not set the timezone if it exists, langset.sh created it
 run rm -f /etc/localtime
--- a/files/usr/lib/jeos-firstboot
+++ b/files/usr/lib/jeos-firstboot
@@ -26,6 +26,10 @@ TEXTDOMAIN='jeos-firstboot'
 . /etc/os-release
 . "$0-functions"
 
+# Read the optional configuration file
+[ -f /usr/share/defaults/jeos-firstboot.conf ] && . /usr/share/defaults/jeos-firstboot.conf
+[ -f /etc/jeos-firstboot.conf ] && . /etc/jeos-firstboot.conf
+
 stty_size() {
     set -- `stty size`; LINES=$1; COLUMNS=$2
     # stty size can return zero when not ready or
--- a/files/usr/lib/jeos-firstboot
+++ b/files/usr/lib/jeos-firstboot
@@ -183,7 +183,7 @@ findlocales()
     [ -n "$list" ]
 }
 
-if [ -z "$LOCALE_PRESET" ]; then
+if [ -z "$JEOS_LOCALE" ]; then
     default="en_US"
     [ -f /etc/locale.conf ] && locale_lang="$(awk -F= '$1 == "LANG" { split($2,fs,"."); print fs[1]; exit }' /etc/locale.conf)"
     [ -n "$locale_lang" ] && default="$locale_lang"
@@ -208,11 +208,11 @@ if [ -z "$LOCALE_PRESET" ]; then
         run langset.sh $locale || warn $"Setting the locale failed"
     fi
 else
-    locale="$LOCALE_PRESET"
+    locale="$JEOS_LOCALE"
     systemd_firstboot_args+=("--locale=$locale")
 fi
 
-if [ -z "$KEYTABLE_PRESET" ]; then
+if [ -z "$JEOS_KEYTABLE" ]; then
     default="us"
     [ -f /etc/vconsole.conf ] && vconsole_keymap="$(awk -F= '$1 == "KEYMAP" { split($2,fs,"."); print fs[1]; exit }' /etc/vconsole.conf)"
     [ -n "$vconsole_keymap" ] && default="$vconsole_keymap"
@@ -245,8 +245,10 @@ if [[ "$(ps h -o tty -p $$)" = tty[0-9]* ]]; then
     start_kmscon["ko"]=1
 
     if [ -n "$locale" -a -n "${start_kmscon[${language}]+_}" ]; then
-        export LOCALE_PRESET="$locale"
-        export KEYTABLE_PRESET="$keytable"
+	# We re-export the variables, as will be inspected when this
+	# script is re-executed in the secondary console
+        export JEOS_LOCALE="$locale"
+        export JEOS_KEYTABLE="$keytable"
 
         if kmscon_available; then
             ret_file="$(mktemp)"
@@ -293,7 +295,7 @@ fi
 default="$(readlink -f /etc/localtime)"
 default="${default##/usr/share/zoneinfo/}"
 
-if [ -z "$TIMEZONE_PRESET" ]; then
+if [ -z "$JEOS_TIMEZONE" ]; then
     # timedatectl doesn't work as dbus is not up yet
     # menulist timedatectl --no-pager list-timezones
     if menulist awk \
@@ -307,7 +309,7 @@ if [ -z "$TIMEZONE_PRESET" ]; then
 	d --msgbox $"error setting timezone" 0 0
     fi
 else
-    timezone="$TIMEZONE_PRESET"
+    timezone="$JEOS_TIMEZONE"
     systemd_firstboot_args+=("--timezone=$timezone")
 fi
 
@@ -315,7 +317,7 @@ fi
 run rm -f /etc/localtime
 run systemd-firstboot "${systemd_firstboot_args[@]}"
 
-if [ -z "$PASSWORD_ALREADY_SET" ]; then
+if [ -z "$JEOS_PASSWORD_ALREADY_SET" ]; then
     # NOTE: must be last as dialog file is used to set the password
     while true; do
 	password=
--- a/files/usr/lib/jeos-firstboot
+++ b/files/usr/lib/jeos-firstboot
@@ -396,7 +396,7 @@ if ! [ -f "$EFI_SYSTAB" ]; then
 fi
 
 # Test if snapper is available
-if [ -x /usr/bin/snapper ]; then
+if [ -x /usr/bin/snapper -a "$(stat --format=%T -f /)" = "btrfs" ]; then
     if ! btrfs qgroup show / &>/dev/null; then
         # Run snapper to setup quota for btrfs
         run /usr/bin/snapper --no-dbus setup-quota || warn $"Could not setup quota for btrfs"
--- a/files/usr/lib/jeos-firstboot
+++ b/files/usr/lib/jeos-firstboot
@@ -311,28 +311,30 @@ fi
 run rm -f /etc/localtime
 run systemd-firstboot "${systemd_firstboot_args[@]}"
 
-# NOTE: must be last as dialog file is used to set the password
-while true; do
-    password=
-    if d --insecure --passwordbox  $"Enter root Password" 0 0; then
-	password="$result"
-	if d --insecure --passwordbox  $"Confirm root Password" 0 0; then
-	    if [ "$password" != "$result" ]; then
-		d --msgbox $"Entered passwords don't match" 5 40 || true
-		continue
+if [ -z "$PASSWORD_ALREADY_SET" ]; then
+    # NOTE: must be last as dialog file is used to set the password
+    while true; do
+	password=
+	if d --insecure --passwordbox  $"Enter root Password" 0 0; then
+	    password="$result"
+	    if d --insecure --passwordbox  $"Confirm root Password" 0 0; then
+		if [ "$password" != "$result" ]; then
+		    d --msgbox $"Entered passwords don't match" 5 40 || true
+		    continue
+		fi
+		# don't use that one as we need to switch locale
+		#systemd_firstboot_args+=("--root-password-file=$dialog_out")
 	    fi
-	    # don't use that one as we need to switch locale
-	    #systemd_firstboot_args+=("--root-password-file=$dialog_out")
 	fi
-    fi
-    if [ -z "$password" ]; then
-	warn $"Warning: No root password set.
+	if [ -z "$password" ]; then
+	    warn $"Warning: No root password set.
 
 You cannot log in that way. A debug shell will be started on tty9 just this time. Use it to e.g. import your ssh key." 0 0 || true
-	run systemctl start debug-shell.service
-    fi
-    break
-done
+	    run systemctl start debug-shell.service
+	fi
+	break
+    done
+fi
 
 if [ -x /usr/bin/SUSEConnect ]; then
     d --msgbox $"Please register this image using your existing SUSE entitlement.
--- a/files/usr/lib/jeos-firstboot
+++ b/files/usr/lib/jeos-firstboot
@@ -104,10 +104,13 @@ sleep 1
 
 systemd_firstboot_args=('--setup-machine-id')
 
+# If the configuration is not loaded and we are in the first terminal
+# instance, make sure that the variables are declared.
+JEOS_LOCALE=${JEOS_LOCALE-}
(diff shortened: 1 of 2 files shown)


files/usr/lib/jeos-firstboot
```diff
@@ -467,7 +474,7 @@ if ! [ -f "$EFI_SYSTAB" ]; then
 fi
 
 # Test if snapper is available
-if [ -x /usr/bin/snapper ]; then
+if [ -x /usr/bin/snapper -a $(stat --format=%T -f /) = "btrfs" ]; then
```
Vogtinator: Please add quotes here:

```suggestion
if [ -x /usr/bin/snapper -a "$(stat --format=%T -f /)" = "btrfs" ]; then
```

aplanas: Indeed. Done

aplanas:
Already tested for my use case.

lnussel:
how do you set those variables? Maybe jeos-firstboot needs a config nowadays and branding packages for different  use cases.

aplanas:
@lnussel creating systemd overlays:

```bash
#======================================
# Overlay for jeos-firstboot (Yomi)
#--------------------------------------
if [[ "$kiwi_profiles" == *"Live"* ]]; then
	LOCALE="en_US"
	KEYTABLE="jp"
	TIMEZONE="UTC"
	mkdir -p /etc/systemd/system/jeos-firstboot.service.d/
	cat > /etc/systemd/system/jeos-firstboot.service.d/10-presets.conf <<-EOF
		[Service]
		Environment=SYSTEMCTL_OPTIONS=--ignore-dependencies TERM=linux LOCALE_PRESET=$LOCALE KEYTABLE_PRESET=$KEYTABLE TIMEZONE_PRESET=$TIMEZONE PASSWORD_ALREADY_SET=1
		ExecStartPost=/usr/sbin/langset.sh $LOCALE $KEYTABLE
	EOF
fi
```

lnussel:
unless you have to meet deadlines so you have to accept hacks I'd recommend to find a better way instead :)

Vogtinator:
What I suggested in #44 is to just add

```
EnvironmentFile=-/usr/share/defaults/jeos-firstboot.conf
EnvironmentFile=-/etc/jeos-firstboot.conf
```

to the .service file and use that for configuration. Alternatively, just sourcing those files inside jeos-firstboot.

aplanas:
@Vogtinator This makes sense to me. I will change the .service, as is a more clear way.

But I wonder if this does not change the current code in this patch.

aplanas:
@Vogtinator if I do a source inside the code, am I breaking the usage of LOCALE_PRESET and KEYTABLE_PRESET?

Vogtinator:
No - if those variables are set in one of the files, it's skipped as it should be. The name of those variables should be changed though to accommodate for the newly gained external visibility.

files/usr/lib/jeos-firstboot
```diff
@@ -280,15 +283,15 @@ if [ -z "$TIMEZONE_PRESET" ]; then
...
+    timezone="$JEOS_TIMEZONE"
     systemd_firstboot_args+=("--timezone=$timezone")
 fi
 
 # systemd-firstboot does not set the timezone if it exists, langset.sh created it
 run rm -f /etc/localtime
 run systemd-firstboot "${systemd_firstboot_args[@]}"
 
-if [ -z "$PASSWORD_ALREADY_SET" ]; then
+if [ -z "$JEOS_PASSWORD" ]; then
```
Vogtinator: `JEOS_PASSWORD_ALREADY_SET`

aplanas: done!

aplanas:
@Vogtinator @lnussel Added the configuration file and renamed the variables. Note that there are two exports for LOCALE_PRESET and KEYTABLE_PRESET that I maintained, as I am not sure if any other process are using them later.

files/usr/lib/jeos-firstboot
```diff
@@ -26,6 +26,9 @@ TEXTDOMAIN='jeos-firstboot'
 . /etc/os-release
 . "$0-functions"
 
+# Read the optional configuration file
+[ -f /etc/jeos-firstboot.conf ] && . /etc/jeos-firstboot.conf
```
Vogtinator: It should either do

```
[ -f /usr/share/defaults/jeos-firstboot.conf ] && . /usr/share/defaults/jeos-firstboot.conf
[ -f /etc/jeos-firstboot.conf ] && . /etc/jeos-firstboot.conf
```

or 

```
if [ -f /etc/jeos-firstboot.conf ]; then
    . /etc/jeos-firstboot.conf
elif [ -f /usr/share/defaults/jeos-firstboot.conf ]; then
    . /usr/share/defaults/jeos-firstboot.conf
fi
```

aplanas: Right. I changed the path of the initial example too.

Done.

lnussel:
&& is not -e safe. Use inverse logic or the if variant.

Vogtinator:
> && is not -e safe. Use inverse logic or the if variant.

It actually is:

```
The
              ERR  trap is not executed if the failed command is part of the command list immediately following a while or until keyword, part of the test in an if statement, part of a command executed in a && or || list except the com-
              mand following the final && or ||, any command in a pipeline but the last, or if the command's return value is being inverted using !.  These are the same conditions obeyed by the errexit (-e) option.
```

aplanas:
I agree with @Vogtinator. Will exit if cannot read the file at the end of the chain (tested locally)

aplanas:
Rename the reexport, as is only used in the second virtual console. Not sure why there is a second virtual console involved, tho.

aplanas:
Partially tested with:

```bash
#======================================
# Config for jeos-firstboot (Yomi)
#--------------------------------------
if [[ "$kiwi_profiles" == *"Live"* ]]; then
	cat > /etc/jeos-firstboot.conf <<-EOF
		JEOS_LOCALE="en_US"
		JEOS_KEYTABLE="jp"
		JEOS_TIMEZONE="UTC"
		JEOS_PASSWORD_ALREADY_SET=1
	EOF
fi
```

Ready for reviews

files/usr/lib/jeos-firstboot
```diff
@@ -181,11 +185,11 @@ if [ -z "$LOCALE_PRESET" ]; then
         run langset.sh $locale || warn $"Setting the locale failed"
     fi
 else
-    locale="$LOCALE_PRESET"
+    locale="$JEOS_LOCALE"
```
Vogtinator: `$locale` can be replaced by `$JEOS_LOCALE` everywhere.

files/usr/lib/jeos-firstboot
```diff
@@ -194,17 +198,20 @@ if [ -z "$KEYTABLE_PRESET" ]; then
...
-
-                # Activate the selected keyboard layout
-                run langset.sh "$locale" "$keytable" || warn $"Setting the keyboard layout failed"
         fi
     else
         d --msgbox $"Error setting keyboard" 0 0
     fi
 else
-    keytable="$result"
+    keytable="$JEOS_KEYTABLE"
```
Vogtinator: `$keytable` can be replaced by `$JEOS_KEYTABLE` everywhere.

files/usr/lib/jeos-firstboot
```diff
@@ -266,45 +275,52 @@ fi
...
+	    && d --default-item "$default" --menu $"Select Time Zone" 0 0 $dh_menu "${list[@]}"; then
 	if [ -n "$result" ]; then
 	    systemd_firstboot_args+=("--timezone=$result")
 	fi
-else
+    else
 	d --msgbox $"error setting timezone" 0 0
+    fi
+else
+    timezone="$JEOS_TIMEZONE"
```
Vogtinator: `$timezone` can be replaced by `$JEOS_TIMEZONE` everywhere.

files/usr/share/defaults/jeos-firstboot.conf
```diff
@@ -0,0 +1,6 @@
+# Example configuration file for jeos-firstboot
+
+# JEOS_LOCALE='en_US'
```
Vogtinator: There should be a line of documentation for every entry here.

files/usr/lib/jeos-firstboot
aplanas: Not sure that this is correct. `$locale` can come from two different paths and later is used in `langset.sh` later on.

files/usr/lib/jeos-firstboot
aplanas: Same, as the keytable can come from the menu or from the environment.

files/usr/lib/jeos-firstboot
aplanas: Done

files/usr/share/defaults/jeos-firstboot.conf
aplanas: Indeed. Done

Vogtinator:
Yes, those variables can come from two sources, but that won't actually matter. If it's set in the configuration file, the user isn't even asked, so it can't be overwritten.

files/usr/lib/jeos-firstboot
aplanas: Ok. The refactor is now in the last PR

files/usr/lib/jeos-firstboot
aplanas: Removed in the last PR

aplanas:
Not very confident of the last PR. I will test it manually, and lets see what openQA says.

files/usr/lib/jeos-firstboot
```diff
@@ -281,7 +276,7 @@ fi
 if [ -e "$EULA_FILE" -a ! -e "${EULA_FILE%/*}/no-acceptance-needed" ]; then
     if [ "$force_english_license" = "0" ]; then
         for i in "${EULA_FILE%.txt}.$locale.txt" \
```
Vogtinator: Missing replacement here

files/usr/lib/jeos-firstboot
```diff
@@ -201,15 +204,14 @@ if [ -z "$JEOS_LOCALE" ]; then
...
+        systemd_firstboot_args+=("--locale=$JEOS_LOCALE")
 
         # Run langset to get consolefont and keymap set up
-        run langset.sh $locale || warn $"Setting the locale failed"
+        run langset.sh $JEOS_LOCALE || warn $"Setting the locale failed"
     fi
 else
-    locale="$JEOS_LOCALE"
-    systemd_firstboot_args+=("--locale=$locale")
+    systemd_firstboot_args+=("--locale=$JEOS_LOCALE")
```
Vogtinator: Should be moved below the fi. The `[ -n "$newlocale" ]` case is never true anyway.

files/usr/lib/jeos-firstboot
aplanas: Ouch. Done.

files/usr/lib/jeos-firstboot
aplanas: How is so? `$newlocale` have always a default value

Vogtinator: Sorry, I meant false.

aplanas: Right! In that case removing the `else` will add two `--locale=$JEOS_LOCALE` into the systemd array. I can think that this will not break anything, but is confusing.

Vogtinator: I meant something like this:

```
if [ -z "$JEOS_LOCALE" ]; then
    default="en_US"
    [ -f /etc/locale.conf ] && locale_lang="$(awk -F= '$1 == "LANG" { split($2,fs,"."); print fs[1]; exit }' /etc/locale.conf)"
    [ -n "$locale_lang" ] && default="$locale_lang"

    list=() # Set by findlocales
    newlocale="$default"
    if ! findlocales; then
        d --msgbox $"No locales found" 0 0
    elif [ "${#list[@]}" -eq 2 ]; then
        newlocale="${list[0]}"
        d --msgbox $"Locale set to $newlocale.\nTo change to a different one, install glibc-locale and use\n'localectl set-locale LANG=ex_AMPLE.UTF-8'." 8 50
    else
        d --default-item "$default" --menu $"Select System Locale" 0 0 $dh_menu "${list[@]}"
        newlocale="${result}"
    fi

    JEOS_LOCALE="${newlocale}.UTF-8"
fi

run langset.sh $JEOS_LOCALE || warn $"Setting the locale failed"
systemd_firstboot_args+=("--locale=$JEOS_LOCALE")
```

aplanas: Very well. I wonder if a similar refactor can be done in the keytable section

files/usr/share/defaults/jeos-firstboot.conf
```diff
@@ -0,0 +1,18 @@
+# Example configuration file for jeos-firstboot
+
+# Valid system locale that will be used in JeOS. If is not set, a
+# dialog box will ask for the computer locale.
```
Vogtinator:
```suggestion
# dialog box will ask for the system locale.
```

files/usr/share/defaults/jeos-firstboot.conf
```diff
@@ -0,0 +1,18 @@
...
+# JEOS_LOCALE='en_US'
+
+# Keyboard layout used after the boot. If is not set, a dialog box
+# will ask for the keyboard layout.
+# JEOS_KEYTABLE='en'
+
+# Local timezone of the node. If is not set, a dialog box will ask for
```
Vogtinator:
```suggestion
# Local timezone of the system. If is not set, a dialog box will ask for
```

files/usr/share/defaults/jeos-firstboot.conf
```diff
@@ -0,0 +1,18 @@
+# Example configuration file for jeos-firstboot
+
+# Valid system locale that will be used in JeOS. If is not set, a
+# dialog box will ask for the computer locale.
+# JEOS_LOCALE='en_US'
+
+# Keyboard layout used after the boot. If is not set, a dialog box
```
Vogtinator: Also during firstboot:

```suggestion
# Keyboard layout used in the system and during jeos-firstboot. If is not set, a dialog box
```

files/usr/share/defaults/jeos-firstboot.conf
```diff
@@ -0,0 +1,18 @@
...
+# will ask for the keyboard layout.
+# JEOS_KEYTABLE='en'
+
+# Local timezone of the node. If is not set, a dialog box will ask for
+# the local timezone.
+# JEOS_TIMEZONE='UTC'
+
+# If is set to a different value than '0', the dialog box for setting
```
Vogtinator: `"0"` will also skip setting the password.

```suggestion
# If is set to a nonempty value, the dialog box for setting
```

files/usr/share/defaults/jeos-firstboot.conf
```diff
@@ -0,0 +1,18 @@
...
+# the initial password for the root user will not be asked. In this
```
Vogtinator:
```suggestion
# the initial password for the root user will be skipped. In this
```

files/usr/share/defaults/jeos-firstboot.conf
aplanas: Done

files/usr/share/defaults/jeos-firstboot.conf
aplanas: Done

files/usr/share/defaults/jeos-firstboot.conf
aplanas: Done

files/usr/share/defaults/jeos-firstboot.conf
aplanas: Done

files/usr/share/defaults/jeos-firstboot.conf
aplanas: True. Done

files/usr/lib/jeos-firstboot
```diff
@@ -289,45 +288,51 @@ fi
...
+	    && d --default-item "$default" --menu $"Select Time Zone" 0 0 $dh_menu "${list[@]}"; then
 	if [ -n "$result" ]; then
 	    systemd_firstboot_args+=("--timezone=$result")
 	fi
-else
+    else
 	d --msgbox $"error setting timezone" 0 0
+    fi
+else
+    systemd_firstboot_args+=("--timezone=$JEOS_TIMEZONE")
```
Vogtinator: Same procedure here

aplanas: Done

Vogtinator:
Found some typos, but will fix them after testing

PR #50, opened 2019-05-15
Add JEOS_MULTINET configuration option

aplanas:
* Configure all interfaces where DHCP answer

* Fix grep stderr output in is_raspberry that happens in non-rpi devices

### DIFF
--- a/files/usr/lib/jeos-firstboot
+++ b/files/usr/lib/jeos-firstboot
@@ -351,7 +351,6 @@ d --infobox $"Collecting network info ..." 3 33
 #
 shopt -s nullglob
 
-candidates=()
 for p in /sys/class/net/* ; do
         test -f "$p" && continue # skip bonding_masters file
 
@@ -359,20 +358,17 @@ for p in /sys/class/net/* ; do
         udevadm info -q all -p "$p" | grep -qs ID_NET_NAME || continue
         d=${p##*/}
 
-        unset IPADDR GATEWAYS
-        eval `wicked test dhcp4 "$d" 2>/dev/null | grep -E "^IPADDR=|^GATEWAYS="`
+        unset IPADDR
+        eval `wicked test dhcp4 "$d" 2>/dev/null | grep -E "^IPADDR="`
         ip link set down "$d" # set link down after probe once done
 
-        test -n "$IPADDR"   && candidates+=("$d")  || continue # ok, track it
-        test -n "$GATEWAYS" && candidates=("$d")  || continue # just use it
-        break
-done
-
-for d in ${candidates[@]} ; do
-        rm -f "/etc/sysconfig/network/ifcfg-$d" || exit 1
-        printf "STARTMODE=auto\nBOOTPROTO=dhcp\n" \
-                > "/etc/sysconfig/network/ifcfg-$d"
-        break # one is enough
+	# Create a configuration file for each interface that provides
+	# an IPADDR
+	if [ -n "$IPADDR" ]; then
+		rm -f "/etc/sysconfig/network/ifcfg-$d" || exit 1
+		printf "STARTMODE=auto\nBOOTPROTO=dhcp\n" \
+			> "/etc/sysconfig/network/ifcfg-$d"
+	fi
 done
 
 config_wireless=false
--- a/files/usr/lib/jeos-firstboot-functions
+++ b/files/usr/lib/jeos-firstboot-functions
@@ -61,7 +61,7 @@ fbiterm_available() {
 }
 
 is_raspberry() {
-     grep -q Raspberry /proc/device-tree/model
+     grep -q Raspberry /proc/device-tree/model 2> /dev/null
 }
 
 # vim: syntax=sh


aplanas:
Note that is on top of the other branch. I will rebase accordingly if the other branch gets merged.

Vogtinator:
Question is whether the pre-MULTINET behaviour was actually useful.

IMO we can just use this by default, I don't see any downside in enabling DHCP for all interfaces. Which interface is first is also not really guaranteed anyway.

aplanas:
@Vogtinator I agree with your comment. I would drop this logic of IPADDRESS vs GATEWAYS, but maybe there were a reason for that.

Vogtinator:
@aplanas: I'd say: Go for it. Then we also don't need a config option.

aplanas:
@Vogtinator done.

files/usr/share/defaults/jeos-firstboot.conf
```diff
@@ -4,3 +4,4 @@
 # JEOS_KEYTABLE='en'
 # JEOS_TIMEZONE='UTC'
 # JEOS_PASSWORD_ALREADY_SET=0
+# JEOS_MULTINET=0
```
Vogtinator: Not needed anymore

files/usr/lib/jeos-firstboot
```diff
@@ -352,16 +352,14 @@ for p in /sys/class/net/* ; do
...
 
-        test -n "$IPADDR"   && candidates+=("$d")  || continue # ok, track it
-        test -n "$GATEWAYS" && candidates=("$d")  || continue # just use it
-        break
+	# Collect the interfaces that provides IPADDR and GATEWAYS
+	[ -n "$IPADDR" -a -n "$GATEWAYS" ] && candidates+=("$d")
 done
 
 for d in ${candidates[@]} ; do
         rm -f "/etc/sysconfig/network/ifcfg-$d" || exit 1
```
Vogtinator: Can be merged into the loop above now and `candidates` removed

Vogtinator:
IMO `GATEWAYS` should also be ignored. Getting an IP address is good enough for DHCP activation.

files/usr/share/defaults/jeos-firstboot.conf
aplanas: Done

files/usr/lib/jeos-firstboot
aplanas: Sure. Done

aplanas:
> IMO `GATEWAYS` should also be ignored. Getting an IP address is good enough for DHCP activation.

I agree. I will change that according to @gmoro comments.

gmoro:
> Question is whether the pre-MULTINET behaviour was actually useful.
> 
> IMO we can just use this by default, I don't see any downside in enabling DHCP for all interfaces. Which interface is first is also not really guaranteed anyway.
> 
> @gmoro: Any reason this was done this IMO overcomplicated way? I'd rather just enable DHCP for all interfaces which get an IP.

This part was actually coded by the maintainer of wicked, and the idea was to use the first configurable network, of course we can rethink it, if we have multiple nets which one are we going to use for the default route? What exactly would be the use case for multiple networks to be configured in the firstboot?

Vogtinator:
> This part was actually coded by the maintainer of wicked, and the idea was to use the first configurable network, of course we can rethink it, if we have multiple nets which one are we going to use for the default route?

Neither of those options are perfect, but potentially having conflicting options for a default route (which shouldn't happen in a well defined environment) might be better than having to set up DHCP for other interfaces manually in case the selection was wrong.

That depends on wicked's behaviour for the conflict situation though, do you have any more info about that?

aplanas:
> > @gmoro: Any reason this was done this IMO overcomplicated way? I'd rather just enable DHCP for all interfaces which get an IP.
> 
> This part was actually coded by the maintainer of wicked, and the idea was to use the first configurable network, of course we can rethink it, if we have multiple nets which one are we going to use for the default route? 

Uhm, I wonder if this is not task for the DHCP administrator to decide, and not for firstboot.

> What exactly would be the use case for multiple networks to be configured in the firstboot?

The use case is simple: derivatives from JeOS can be used in scenarios were multiple interfaces are available. In cloud and kubernetes cloud-init or ignition will configure the interfaces, but in other environments firstboot will need to assign some same defaults.

files/usr/lib/jeos-firstboot
```diff
@@ -352,16 +351,13 @@ for p in /sys/class/net/* ; do
...
-done
-
-for d in ${candidates[@]} ; do
-        rm -f "/etc/sysconfig/network/ifcfg-$d" || exit 1
-        printf "STARTMODE=auto\nBOOTPROTO=dhcp\n" \
-                > "/etc/sysconfig/network/ifcfg-$d"
-        break # one is enough
+	# Create a configuration file for each interface that provides
+	# IPADDR and GATEWAYS
+	if [ -n "$IPADDR" -a -n "$GATEWAYS" ]; then
```
Vogtinator: Remove the check for `GATEWAYS` here, interfaces without routes might still be useful.

aplanas: Done. Note that for cleaning I also removed the `unset` and the `eval` parts.

Vogtinator:
LGTM, just needs to be rebased on master after the other PR gets merged

aplanas:
Rebased on top of the other PR (that is now also rebased from the new master)

gmoro:
> That depends on wicked's behaviour for the conflict situation though, do you have any more info about that?

I don't, but let's go with it as it is.

Vogtinator:
Don't merge before #49 got merged and this PR rebased on top.

PR #51, opened 2019-05-21
Fix some issues in the recently merged PRs

Vogtinator:
Mostly harmless.


Vogtinator:
Actually, let's merge and submit now as the current state is unusable while this is confirmed to work on my systems already...

PR #52, opened 2019-06-24
Do not fail when updating bootloader in no-UEFI

aplanas:
In non-UEFI JeOS images not always /etc/sysconfig/bootloader is
present. When this system is missing, jeos-firstboot service fail
and shutdown the full machine.

### DIFF
--- a/files/usr/lib/jeos-firstboot
+++ b/files/usr/lib/jeos-firstboot
@@ -383,7 +383,9 @@ EFI_SYSTAB="/sys/firmware/efi/systab"
 # Look for EFI dir to see if the machine is booted in UEFI mode
 run modprobe efivars
 if ! [ -f "$EFI_SYSTAB" ]; then
-	run sed -i -e "s/LOADER_TYPE=.*/LOADER_TYPE=grub2/g" /etc/sysconfig/bootloader
+	if [ -f /etc/sysconfig/bootloader ]; then
+		run sed -i -e "s/LOADER_TYPE=.*/LOADER_TYPE=grub2/g" /etc/sysconfig/bootloader
+	fi
 fi
 
 # Test if snapper is available


Vogtinator:
That's weird - kiwi should write that unconditionally: https://github.com/SUSE/kiwi/blob/4d0c77ba1c58914a4eaa6e3613a587dd32c2e288/kiwi/bootloader/config/grub2.py#L203

Do you have the mentioned log string in the build log?

aplanas:
Can be something non-usual in my KIWI file, as I do not have it in my logs:

https://build.opensuse.org/build/home:aplanas:Images/images/x86_64/openSUSE-Tumbleweed-JeOS:Live/_log

This is my XML, as a Live image.

https://build.opensuse.org/package/rdiff/home:aplanas:Images/openSUSE-Tumbleweed-JeOS?opackage=openSUSE-Tumbleweed-JeOS&oproject=Virtualization%3AAppliances%3AImages%3AopenSUSE-Tumbleweed&rev=31

Vogtinator:
It's actually in there:

```
[  830s] [ INFO    ]: 14:07:25 | Writing sysconfig bootloader file
```

So it appears like something's deleting the file again.

aplanas:
Ouch right.

I wonder what is removing this file, as the config.sh is only adding assets.

aplanas:
Actually, can be that nothing is written, and there is not entries for that in the logs:

https://github.com/SUSE/kiwi/blob/master/kiwi/bootloader/config/grub2.py#L218

Vogtinator:
KIWI always sets entries though:

https://github.com/SUSE/kiwi/blob/master/kiwi/bootloader/config/grub2.py#L188

```
        sysconfig_bootloader_entries = {
            'LOADER_TYPE':
                'grub2-efi' if self.firmware.efi_mode() else 'grub2',
            'LOADER_LOCATION':
                'none' if self.firmware.efi_mode() else 'mbr'
        }
```

so something's broken with kiwi as well?

aplanas:
> so something's broken with kiwi as well?

I think so : (

The other way to skip writing is that the directory `/etc/sysconfig/` is not yet present

aplanas:
I debugged a bit Kiwi, and I created https://github.com/SUSE/kiwi/issues/1112. But is the check `os.path.exists(sysconfig_bootloader_location)` suggest that calling this method when the `squashfs.img` is already created and there is no `/etc/sysconfig/` anymore is expected by Kiwi.

If so, the check in jeos-firstboot is required.

Vogtinator:
IMO it's a bug in kiwi that needs to be fixed, but the code change here makes sense independently.

PR #55, opened 2019-08-08
Run jeos-firstboot-snapshot after local-fs and ignore systemd-user-sessions

Vogtinator:
Failed in weird ways in openQA, "multi-user.target not reached" without any hint as to why.

With this it appears to work, at least once.


Vogtinator:
One successful run in openQA and works locally as well

PR #58, opened 2019-10-08
Ignore udevadm and modprobe errors for use in containers

### LINKED ISSUE #56: jeos-firstboot calls modprobe in "containers"
running jeos-firstboot eg in WSL it still tries to call modprobe which doesn't work of course. Editing /etc/sysconfig/bootloader seems like a rather dirty hack anyways. Maybe should be moved to a hook that can be disabled instead,

### LINKED ISSUE #57: network settings are unconditional
the network detection and configuration is unconditional. It may not be needed if the image already has network configured or there is nothing to do actually, like in WSL

Vogtinator:
- Fixes #56
- Fixes #57

### DIFF
--- a/files/usr/lib/jeos-firstboot
+++ b/files/usr/lib/jeos-firstboot
@@ -351,8 +351,9 @@ shopt -s nullglob
 for p in /sys/class/net/* ; do
 	test -f "$p" && continue # skip bonding_masters file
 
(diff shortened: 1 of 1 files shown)


Vogtinator:
I added some comments. Does it work?

lnussel:
yes works

PR #59, opened 2019-10-10
SUSEConnect: do not show the register dialog on non-SLE distributions

aplanas:
In openSUSE based distributions we do not want to show the SUSEConnect dialog when this package is available, as the updates are not related with registration.

For installation media, installing SUSEConnect in openSUSE will make sense.

### DIFF
--- a/files/usr/lib/jeos-firstboot
+++ b/files/usr/lib/jeos-firstboot
@@ -330,7 +330,8 @@ You cannot log in that way. A debug shell will be started on tty9 just this time
 	done
 fi
 
-if [ -x /usr/bin/SUSEConnect ]; then
+# Do not show the register on non SLE based distributions
+if [ -x /usr/bin/SUSEConnect -a -z "${ID##sle*}" ]; then
 	d --msgbox $"Please register this image using your existing SUSE entitlement.
 
 As \"root\" use the following command:


Vogtinator:
> For installation media, installing SUSEConnect in openSUSE will make sense.

How?

aplanas:
> > For installation media, installing SUSEConnect in openSUSE will make sense.
> 
> How?

For one side, SUSEConnect is in Factory, so we can create images that install it and use SUSEConnet, just only in this case the dialog is missleading.

Why install SUSEConnect package in TW based image that uses jeos-firstboot?

My current use case is Yomi: `SUSEConnect --root /mnt ...` will allow the registration of an SLE based system that is living in the chroot, from an openSUSE based image that is living in RAM.

aplanas:
> Instead of "not on openSUSE" it should check for SLE. Something like `-n "${ID##*sle*}"` might work.

Done. Scares me show small is the regex, tho.

Vogtinator:
> Is product.xml in a livecd image? I think that is only on real installation media.

Yup, but `/etc/products.d/baseproduct` is installed and seems to have some helpful data inside...

aplanas:
Changed the regex for one a bit more precise

Vogtinator:
Looks like it's doing the opposite now, showing the registration only on non-sle

aplanas:
> Looks like it's doing the opposite now, showing the registration only on non-sle

No, is correct:

```bash
# cat /etc/os-release 
NAME="SLES"
VERSION="12-SP4"
VERSION_ID="12.4"
PRETTY_NAME="SUSE Linux Enterprise Server 12 SP4"
ID="sles"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:suse:sles:12:sp4"

# ls /usr/bin/SUSEConnect
/usr/bin/SUSEConnect

# if [ -x /usr/bin/SUSEConnect -a -z "${ID##sle*}" ]; then echo 'IN SLE'; fi
IN SLE

# if [ -x /usr/bin/SUSEConnect -a -n "${ID##sle*}" ]; then echo 'IN SLE'; fi
```

PR #61, opened 2019-11-27
Improve Rpi wireless config, jsc#SLE-10603

gmoro:
Improve Rpi wireless config, jsc#SLE-10603

Implements ability to repeat configuration in case of failure

### DIFF
--- a/files/usr/lib/jeos-firstboot
+++ b/files/usr/lib/jeos-firstboot
@@ -122,8 +122,11 @@ d(){
 	retval=
 	while true
 	do
+        # disable exit on error for capturing dialog return codes
+	        set +e
 		dialog --backtitle "$PRETTY_NAME" --output-fd 3 "$@" 3>"${dialog_out}"
 		retval=$?
+        	set -e
 		case $retval in
 		  0)
 			# need || true as dialog doesn't write newlines
--- a/files/usr/share/jeos-firstboot/raspberrywifi
+++ b/files/usr/share/jeos-firstboot/raspberrywifi
@@ -75,7 +75,7 @@ raspberrywifi_systemd_firstboot()
 		if [ "${#list[@]}" -eq "2" ]; then
 			wlan_device="${list[0]}"
 		else
-			d --menu  $"Select Wireless card to configure" 0 0 $dh_menu "${list[@]}"
+			d --menu  $"Select wireless card to configure" 0 0 $dh_menu "${list[@]}"
 			wlan_device="${result}"
 		fi
 
@@ -105,15 +105,13 @@ raspberrywifi_systemd_firstboot()
 		fi
 		if [ "$wlan_auth_mode" != "Open" ]; then
 			wlan_password=
-			d --insecure --passwordbox  $"Password" 0 0
+			d --insecure --passwordbox  $"Network password" 0 0
 			wlan_password="$result"
 		fi
-		break
-	done
 
-	config_file=`mktemp -qt 'firstboot-XXXXXX'`
+		config_file=`mktemp -qt 'firstboot-XXXXXX'`
 
-	cat << EOF > $config_file
+		cat << EOF > $config_file
 BOOTPROTO='dhcp'
 STARTMODE='auto'
 WIRELESS_AP_SCANMODE='1'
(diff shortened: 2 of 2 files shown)


files/usr/lib/jeos-firstboot
```diff
@@ -122,8 +122,11 @@ d(){
 	retval=
 	while true
 	do
+        # disable exit on error for capturing dialog return codes
+	        set +e
```
Vogtinator:
```
retval=0
dialog --backtitle "$PRETTY_NAME" --output-fd 3 "$@" 3>"${dialog_out}" || retval=$?
```

files/usr/share/jeos-firstboot/raspberrywifi
```diff
@@ -122,23 +120,40 @@ WIRELESS_ESSID='$wlan_network'
...
+			echo "WIRELESS_EAP_AUTH='mschapv2'" >> $config_file
+		fi
 
-	run mv -f $config_file /etc/sysconfig/network/ifcfg-$wlan_device
+		run mv -f $config_file /etc/sysconfig/network/ifcfg-$wlan_device
+		d --infobox $"Connecting to wireless network ..." 3 38 || true
+
+		run ifdown $wlan_device &>/dev/null || true
+		# Disabling exit on error to capture exit code
+		set +e
```
Vogtinator: Bash 101:

```
if ifup $wlan_device &>/dev/null; then
...
fi
```

files/usr/share/jeos-firstboot/raspberrywifi
```diff
@@ -122,23 +120,40 @@ WIRELESS_ESSID='$wlan_network'
...
+		ifup $wlan_device &>/dev/null
+		ret_val=$?
+		set -e
+		if [ $ret_val -ne 0 ]; then
+			if dialog --yesno $"Connection failed, do you wish to retry?" 0 0; then
```
Vogtinator: Is it possible to get an error message from wicked? This error message is pretty useless

files/usr/share/jeos-firstboot/raspberrywifi
```diff
@@ -122,23 +120,40 @@ WIRELESS_ESSID='$wlan_network'
...
+				continue
+			else
+				# Otherwise, return 0 to continue main execution
+				return 0
```
Vogtinator: Please use `break` and `return 0` consistently

files/usr/share/jeos-firstboot/raspberrywifi
```diff
@@ -122,23 +120,40 @@ WIRELESS_ESSID='$wlan_network'
...
+			fi
+		else
+			break
+		fi
+	done
 }
 
 # This is called after the configuration steps finished successfully
 raspberrywifi_post()
```
Vogtinator: Just remove this?

files/usr/lib/jeos-firstboot
gmoro: Stylistic choice right?

files/usr/share/jeos-firstboot/raspberrywifi
gmoro: again, it's broke down the way it is on purpose, to make it clearer,  pure stylistic choice.

files/usr/share/jeos-firstboot/raspberrywifi
gmoro: Sadly no, wicked does not have more then a setup-in-progress message, and wpa_supplicant uses a D-bus api to give status, will be more complicated, but I will check on how to get proper errors later.

files/usr/share/jeos-firstboot/raspberrywifi
gmoro: I do not understand what is inconsistent here, can you please clarify?

files/usr/share/jeos-firstboot/raspberrywifi
gmoro: I think is best to leave it so this code can serve as a model for any other "module" for jeos-firstboot, I can actually document the mechanism as well, put some comments maybe.

files/usr/share/jeos-firstboot/raspberrywifi
Vogtinator: No, this is simply bad code.

files/usr/lib/jeos-firstboot
Vogtinator: Could be argued, but changing the shell mode should be avoided.

files/usr/share/jeos-firstboot/raspberrywifi
Vogtinator: You use `break` below to achieve the same.

files/usr/share/jeos-firstboot/raspberrywifi
Vogtinator: That should be documented on https://github.com/openSUSE/jeos-firstboot/wiki/Support-for-extensions-using-custom-scripts

PR #62, opened 2020-01-21
Gmoro

gmoro:
Fix bsc#1161068
Fix coding style and some UI parts

### DIFF
--- a/files/usr/lib/jeos-firstboot
+++ b/files/usr/lib/jeos-firstboot
@@ -351,26 +351,26 @@ d --infobox $"Collecting network info ..." 3 33
 #
 shopt -s nullglob
 
-for p in /sys/class/net/* ; do
-	test -f "$p" && continue # skip bonding_masters file
+for net_path in /sys/class/net/* ; do
+	test -f "$net_path" && continue # skip bonding_masters file
 
 	# Only devices having ID_NET_NAME.* attrs
 	# Ignore errors if udev not available
-	udevadm info -q property -p "$p" 2>/dev/null | grep -qs ID_NET_NAME || continue
+	udevadm info -q property -p "$net_path" 2>/dev/null | grep -qs ID_NET_NAME || continue
 	# But don't touch WLAN interfaces
-	udevadm info -q property -p "$p" | grep -qs "DEVTYPE=wlan" && continue
+	udevadm info -q property -p "$net_path" | grep -qs "DEVTYPE=wlan" && continue
 
-	d=${p##*/}
+	net_device=${net_path##*/}
 
 	unset IPADDR
-	eval `wicked test dhcp4 "$d" 2>/dev/null | grep -E "^IPADDR="`
-	ip link set down "$d" # set link down after probe once done
+	eval `wicked test dhcp4 "$net_device" 2>/dev/null | grep -E "^IPADDR="`
+	ip link set down "$net_device" # set link down after probe once done
 
 	# Create a configuration file for each interface that provides
 	# an IPADDR
 	if [ -n "$IPADDR" ]; then
 		printf "STARTMODE=auto\nBOOTPROTO=dhcp\n" \
-			> "/etc/sysconfig/network/ifcfg-$d"
+			> "/etc/sysconfig/network/ifcfg-$net_device"
 	fi
 done
 
--- a/files/usr/lib/jeos-firstboot
+++ b/files/usr/lib/jeos-firstboot
@@ -204,7 +204,7 @@ if [ -z "$JEOS_LOCALE" ]; then
 		newlocale="${list[0]}"
 		d --msgbox $"Locale set to $newlocale.\nTo change to a different one, install glibc-locale and use\n'localectl set-locale LANG=ex_AMPLE.UTF-8'." 8 50
 	else
-		d --default-item "$default" --menu $"Select System Locale" 0 0 $dh_menu "${list[@]}"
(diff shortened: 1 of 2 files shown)


files/usr/lib/jeos-firstboot
```diff
@@ -122,11 +122,9 @@ d(){
 	retval=
 	while true
 	do
-        # disable exit on error for capturing dialog return codes
-	        set +e
-		dialog --backtitle "$PRETTY_NAME" --output-fd 3 "$@" 3>"${dialog_out}"
-		retval=$?
-        	set -e
+	        # Note: Remember about the set -e if changing this line
```
Vogtinator: That pretty much applies to the whole file, so IMO that comment is misplaced/can be removed

files/usr/share/jeos-firstboot/raspberrywifi
```diff
@@ -56,7 +56,24 @@ is_raspberry()
...
+                # Only devices having ID_NET_NAME.* attrs
+                # Ignore errors if udev not available
+                udevadm info -q property -p "$p" 2>/dev/null | grep -qs ID_NET_NAME || continue
+                # But don't touch WLAN interfaces
+                udevadm info -q property -p "$p" | grep -qs "DEVTYPE=wlan" && continue
+
+                d=${p##*/}
+
+                # On Rpi create a configuration file for each interface detected
+                # even when no cable connected
```
Vogtinator: That needs some discussion:

* Runs even on non-RPi. The `is_raspberry` check is below.
* Why just for the RPi? It's an unexpected difference
* It can be a (negative) surprise in general to have this configured for an unused interface. There should at least be a dialog "DHCP enabled for these interfaces: eth0"

files/usr/share/jeos-firstboot/raspberrywifi
```diff
@@ -56,7 +56,24 @@ is_raspberry()
 # interaction and access to the global systemd_firstboot_args array
 raspberrywifi_systemd_firstboot()
 {
-	if is_raspberry && ls -d /sys/class/net/*/wireless &>/dev/null; then
+        for p in /sys/class/net/* ; do
```
Vogtinator: `p` and `d` aren't great variable names

files/usr/lib/jeos-firstboot
```diff
@@ -286,8 +284,8 @@ if [ -e "$EULA_FILE" -a ! -e "${EULA_FILE%/*}/no-acceptance-needed" ]; then
 		done
 	fi
 
-	while ! dialog --backtitle "$PRETTY_NAME" --textbox "$EULA_FILE" $dh_text 85 --and-widget --yesno $"Do you agree with the terms of the license?" 0 0; do
-		d --msgbox $"Well, we cannot continue then ..." 6 40
+	while ! d --backtitle "$PRETTY_NAME" --textbox "$EULA_FILE" $dh_text 85 --and-widget --yesno $"Do you agree with the terms of the license?" 0 0; do
+		true
```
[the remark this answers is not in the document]
gmoro: Agreed, will change to a message and give the option to shutdown

files/usr/share/jeos-firstboot/raspberrywifi
```diff
@@ -56,7 +56,24 @@ is_raspberry()
...
+
+                # Only devices having ID_NET_NAME.* attrs
+                # Ignore errors if udev not available
+                udevadm info -q property -p "$p" 2>/dev/null | grep -qs ID_NET_NAME || continue
+                # But don't touch WLAN interfaces
+                udevadm info -q property -p "$p" | grep -qs "DEVTYPE=wlan" && continue
+
+                d=${p##*/}
+
+                # On Rpi create a configuration file for each interface detected
```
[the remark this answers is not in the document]
gmoro: good catch

files/usr/share/jeos-firstboot/raspberrywifi
gmoro: Yeah, let's discuss that later then.

files/usr/share/jeos-firstboot/raspberrywifi
gmoro: just copied and pasted from the code on /lib/jeos-firstboot, gonna change in both places.

files/usr/share/jeos-firstboot/raspberrywifi
gmoro: I changed to have all interfaces configured with DHCP, at least while we don't have a way to re-run the network configuration standalone, and just for RPi as if applied to the virtual ones would be a difference in behaviour (currently we configure just the first one connected).

files/usr/lib/jeos-firstboot
```diff
@@ -268,13 +265,13 @@ fi
...
-# Failsafe: If neither a license nor the no-acceptance-needed flag are found, quit.
-if ! [ -e "$EULA_FILE" -o -e "${EULA_FILE%/*}/no-acceptance-needed" ]; then
+# Failsafe: If no license found, quit.
+if ! [ -e "$EULA_FILE" ]; then
 	d --msgbox $"No license found - cannot continue" 6 40
 	exit 1
 fi
 
-if [ -e "$EULA_FILE" -a ! -e "${EULA_FILE%/*}/no-acceptance-needed" ]; then
+if [ -e "$EULA_FILE" ]; then
```
Vogtinator: This condition is now always true.

PR #66, opened 2020-03-26
Add JEOS_HIDE_SUSECONNECT option

### LINKED ISSUE #65: Add an option to hide the SUSEConnect dialog on SLE
When developing a PXE boot image based on SLE, we may want to hide the SUSEConnect dialog to avoid any user interaction.

aplanas:
In SLE based images where SUSEConnect is available, a dialog is shown to
the user to remember the registration of the product in order to fetch
future updates.

This commit add a new option that, if enabled, will hide the dialog and
avoid this user interaction.

Fix #65

### DIFF
--- a/files/usr/share/defaults/jeos-firstboot.conf
+++ b/files/usr/share/defaults/jeos-firstboot.conf
@@ -22,3 +22,8 @@
 # PXE / OEM) that are part of another product and than can be used
 # when this license was already accepted by other means.
 # JEOS_EULA_ALREADY_AGREED='yes'
+
+# If set to a nonempty value, the dialog box for showing the
+# SUSEConnect help will not be displayed. By default this dialog is
+# present when SUSEConnect is installed on SLE systems.
+JEOS_HIDE_SUSECONNECT='yes'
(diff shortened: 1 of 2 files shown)


files/usr/share/defaults/jeos-firstboot.conf
```diff
@@ -22,3 +22,9 @@
 # PXE / OEM) that are part of another product and than can be used
 # when this license was already accepted by other means.
 # JEOS_EULA_ALREADY_AGREED='yes'
+
+# If set to a nonempty value, the dialog box for showing the
```
Vogtinator: Has `dialog` twice

files/usr/share/defaults/jeos-firstboot.conf
```diff
@@ -22,3 +22,9 @@
...
+# SUSEConnect dialog help will not be displayed. By default this
+# dialog is present when SUSEConnect is installed on SLE systems, and
```
Vogtinator: I'd just drop the `, and requires user intervention to hide it.` - that's just normal dialog behaviour

aplanas: Yes, done.

files/usr/share/defaults/jeos-firstboot.conf
aplanas: Fixed.

PR #69, opened 2020-05-08
Move scripts to /usr/sbin

### LINKED ISSUE #67: Move to /usr/libexec
`%{_libexecdir}` will be changed to `/usr/libexec` for Tumbleweed, so the build fails:

```
[   13s] RPM build errors:
[   13s]     File not found: /home/abuild/rpmbuild/BUILDROOT/jeos-firstboot-0.0+git20200331.db3bc7c-0.x86_64/usr/libexec/jeos-firstboot
[   13s]     File not found: /home/abuild/rpmbuild/BUILDROOT/jeos-firstboot-0.0+git20200331.db3bc7c-0.x86_64/usr/libexec/jeos-firstboot-snapshot
[   13s]     File not found: /home/abuild/rpmbuild/BUILDROOT/jeos-firstboot-0.0+git20200331.db3bc7c-0.x86_64/usr/libexec/jeos-firstboot-functions
```

We should move to `/usr/libexec` and also merge `jeos-firstboot-functions` into `jeos-firstboot`.

How should we deal with current TW and Leap/SLE? Just `mkdir /usr/libexec` and use the new location there as well or do some `mv`/`sed` to choose the location depending on the OS version?

Vogtinator:
Fixes #67


gmoro:
Why not take the opportunity and just move this to /usr/bin instead? I don't see a reason to the shell script to live under /usr/lib anyway

Vogtinator:
Because it doesn't really make sense to have them in $PATH.

Vogtinator:
Though this would elegantly avoid the issue with distros without `/usr/libexec`, so I'd actually be in favor of that now...

gmoro:
> Because it doesn't really make sense to have them in $PATH.
well, you can run it after the firstboot if needed, maybe move to /usr/sbin instead, to avoid having it in the regular user $PATH

PR #70, opened 2020-08-27
Drop the info dialog about glibc-locale (jsc#SLE-15348)

Vogtinator:
This belongs in the documentation instead, which saves one mandatory
interaction on the first boot.


Vogtinator:
Simple change, tested successfully on Tumbleweed.

PR #71, opened 2020-09-30
Use locale en_US.UTF-8 by default

aplanas:
Locale en_US do not exist anymore, and firstboot complain about this,
generating an exit and shuttind down the boot process.


Vogtinator (changes requested):
In the "No locales found" case it would result in `JEOS_LOCALE="en_US.UTF-8.UTF-8"`.

What does actually fail here? `UTF-8` is intentionally omitted everywhere and added in line 206.

PR #74, opened 2020-11-24
Refactor

gmoro:
Major refactor of the code base

Will use this as base for new scheme of branched and tagged releases from now on


files/usr/sbin/jeos-config
```diff
@@ -0,0 +1,147 @@
...
+subcommand=$1; shift  # Remove 'pip' from the argument lis
+
+case "$subcommand" in
+
+	locale)
+		dialog_locale
+		apply_locale
+        	;;
+	keytable)
+		dialog_locale
```
Vogtinator, line 132: This should just keep the selected locale instead of asking again

Vogtinator (changes requested):
Please run shellcheck and fix the reported errors. Indentation has to be fixed to be consistent.

`jeos-firstboot-functions` needs to be cleaned up, currently it's just a place where all kinds of stuff is in. Separating the "built-in" dialogs into module files might work.

Now there's a separate module hook needed for supporting `jeos-config`, but modules without that would still be shown in the list, but fail to run. Not sure what to do about that though.

files/usr/sbin/jeos-config
```diff
@@ -0,0 +1,147 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+. /etc/os-release
+. "/usr/share/jeos-firstboot/jeos-firstboot-functions"
```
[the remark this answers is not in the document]
gmoro: Aware of that, but just executed if has the appropriate ${hook} function, anyway, need to improve that, any ideas? Following FHS should be in share, my initial idea was to move the modules to a separate subfolder called modules, what do you think?

files/usr/sbin/jeos-config
```diff
@@ -0,0 +1,147 @@
...
+JEOS_KEYTABLE=${JEOS_KEYTABLE-}
+
+result=
+list=
+password=''
+
+let dh_menu=LINES-15
+let dh_text=LINES-5
+
+function usage()
```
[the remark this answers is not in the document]
gmoro: correct

files/usr/sbin/jeos-config
```diff
@@ -0,0 +1,147 @@
...
+	for module in *; do
+		if [ -f "${module}" ] && source "${module}"; then
+			modules+=("${module}")
+		fi
+	done
+	popd &>/dev/null
+fi
+
+cleanup() {
+	#call_module_hook cleanup
```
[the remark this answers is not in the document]
gmoro: The config modules still don't have a cleanup function, left there to remember to check if it's necessary, will take a look.

files/usr/sbin/jeos-config
```diff
@@ -0,0 +1,147 @@
...
+	echo "	keytable	Show dialog to configure the system password"
+	echo "	network		Show dialog to configure the system network"
+	for module in "${modules[@]}"; do
+		echo "	${module}	Show dialog of module ${module}"
+	done
+	echo ""
+}
+
+if [ $# -eq 0 ]; then
+	usage
```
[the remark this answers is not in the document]
gmoro: That's a good idea, keep the graphic theming going on!

files/usr/sbin/jeos-config
```diff
@@ -0,0 +1,147 @@
...
+
+let dh_menu=LINES-15
+let dh_text=LINES-5
+
+function usage()
+{
+	echo "Usage: jeos-config [OPTION...] CONFIG_NAME"
+	echo ""
+	echo "	-h		shows these usage help"
+	echo "	locale		Show dialog to configure the system locale"
```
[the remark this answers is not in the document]
gmoro: agreed

files/usr/sbin/jeos-config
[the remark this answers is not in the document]
Vogtinator: `jeos-firstboot` and `jeos-config` are very similar anyway, and so it might even make sense to merge them into one file and thus not need `jeos-firstboot-functions`. A separate `modules/` subdirectory would work as well, I don't think we have to be backwards-compatible there.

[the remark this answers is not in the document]
gmoro: Yes, thought about that as well, kept simple at first, but if we get too much overlap means some more of the code can go in the shared library -functions.

files/usr/share/jeos-firstboot/jeos-firstboot-dialogs
```diff
@@ -0,0 +1,118 @@
+menulist()
+{
+	list=()
+	local line
+	while read line; do
+		list+=("$line" '')
+	done < <("$@"||true)
```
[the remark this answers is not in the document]
gmoro: Not sure, copy and paste from the old code

files/usr/sbin/jeos-config
```diff
@@ -0,0 +1,147 @@
...
+     			exit 1
+			;;
+	esac
+done
+
+subcommand=$1; shift  # Remove 'pip' from the argument lis
+
+case "$subcommand" in
+
+	locale)
```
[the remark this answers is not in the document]
gmoro: Will leave this for the next step, for all this to be turned in modules jeos-firstboot would have to understand ordering of execution of the modules, not sure how I'm gonna implement that yet. There's a separate card in trello for that already.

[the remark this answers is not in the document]
Vogtinator: That's already implemented. The execution is ordered alphabetically.

files/usr/share/jeos-firstboot/jeos-firstboot-functions
```diff
@@ -0,0 +1,121 @@
...
+
+fbiterm_available() {
+	# fbiterm itself is installed
+	fbiterm --help >/dev/null 2>&1 || return 1
+	# fbiterm comes with its own fallback font
+
+	return 0
+}
+
+dialog_out=`mktemp -qt 'firstboot-XXXXXX'`
```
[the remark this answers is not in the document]
gmoro: This didn't really work here, what exactly is it doing, not sure I understand it.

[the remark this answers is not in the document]
Vogtinator: It creates a file, references it and deletes the name on the file system. How does it fail?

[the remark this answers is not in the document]
gmoro: I just says the /proc file descriptor does not exist
I will leave this one as is for now, will take note as a bugfix after we merge this one.

files/usr/sbin/jeos-config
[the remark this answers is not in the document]
gmoro: Will note it down and rework after this merge

files/usr/sbin/jeos-config
```diff
@@ -0,0 +1,147 @@
...
+	run() {
+		echo "$@"
+	}
+else
+	run() {
+		"$@"
+	}
+fi
+
+call_module_config() {
```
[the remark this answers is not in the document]
gmoro: moved

files/usr/sbin/jeos-config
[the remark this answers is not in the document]
gmoro: Alphabetical is note enough for now, as we still use the file names as names so adding 01-module-name will not exactly be enough, anyway, there's a trello card and we can rework that later.

files/usr/share/jeos-firstboot/jeos-firstboot-functions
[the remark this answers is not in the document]
gmoro: Card created on trello for left over fixes.
https://trello.com/c/Y2IJ8GXX/354-non-critical-fixes-for-jeos-firstboot

files/usr/sbin/jeos-config
```diff
@@ -0,0 +1,147 @@
...
+
+if [ $# -eq 0 ]; then
+	select_config
+	subcommand=${result}
+else
+	subcommand=$1; shift
+fi
+
+case "$subcommand" in
+
```
Vogtinator, line 122: Empty line

files/usr/share/jeos-firstboot/jeos-firstboot-dialogs
```diff
@@ -0,0 +1,118 @@
...
+{
+	default="$(readlink -f /etc/localtime)"
+	default="${default##/usr/share/zoneinfo/}"
+	# timedatectl doesn't work as dbus is not up yet
+	# menulist timedatectl --no-pager list-timezones
+	if menulist awk \
+		'BEGIN{print "UTC"; sort="sort"}/^#/{next;}{print $3|sort}END{close(sort)}' \
+		/usr/share/zoneinfo/zone.tab \
+		&& d --default-item "$default" --menu $"Select time zone" 0 0 $dh_menu "${list[@]}"; then
+	if [ -n "$result" ]; then
```
Vogtinator, line 58: As you touch this anyway, please fix the indentation

files/usr/sbin/jeos-config
```diff
@@ -0,0 +1,147 @@
...
+		dialog_network
+		d --infobox $"Restarting network ..." 3 26 || true
+		systemctl restart network
+		;;
+	password)
+		dialog_password
+		apply_password
+		;;
+	*)
+		call_module_config "$subcommand"|| echo "Unknown option"
```
Vogtinator, line 146: We need a generic `call_module` which is just `call_module_hook` for a single module, then this can be just

`call_module "$subcommand" jeos_config`

files/usr/share/jeos-firstboot/jeos-firstboot-functions
```diff
@@ -0,0 +1,136 @@
...
+			;;
+		esac
+	done
+}
+
+warn(){
+	d --title $"Warning" --msgbox "$1" 6 40
+}
+
+# localectl --no-pager list-keymaps does not list aliases (symlinks), but those are used
```
Vogtinator, line 84: All of the functions below are specific to the functions in `jeos-firstboot-dialogs`, so should be moved there

files/usr/share/jeos-firstboot/jeos-firstboot-functions
```diff
@@ -0,0 +1,136 @@
...
+                locale="${l#/usr/share/langset/}"
+                [ -d "/usr/lib/locale/${locale}.utf8" ] || continue
+                list+=("${locale}" '')
+        done
+        [ -n "$list" ]
+}
+
+get_current_locale()
+{
+	JEOS_LOCALE=`awk -F= '$1 == "LANG" { print $2; exit }' /etc/locale.conf`
```
Vogtinator, line 111: This is incomplete and should fall back to `en_US` if empty. It should also echo the value instead of setting a global variable, that makes it much easier to use.

files/usr/share/jeos-firstboot/jeos-firstboot-dialogs
```diff
@@ -0,0 +1,118 @@
...
+	local line
+	while read line; do
+		list+=("$line" '')
+	done < <("$@"||true)
+	[ -n "$list" ]
+}
+
+dialog_locale()
+{
+        default="en_US"
```
Vogtinator, line 14: After `get_current_locale` is fixed, this can be just `default="$(get_current_locale)"`

files/usr/sbin/jeos-config
```diff
@@ -0,0 +1,147 @@
...
+	run() {
+		echo "$@"
+	}
+else
+	run() {
+		"$@"
+	}
+fi
+
+if pushd "/usr/share/jeos-firstboot/modules" &>/dev/null; then
```
Vogtinator, line 48: This belongs into `jeos-firstboot-functions`.

files/usr/sbin/jeos-config
```diff
@@ -0,0 +1,147 @@
...
+	#call_module_hook cleanup
+	echo .oOo.oOo.oOo. > $dialog_out
+	rm -f "$dialog_out"
+	# reenable systemd and kernel logs
+	run kill -s SIGRTMAX-10 1
+	run setterm -msg on 2>/dev/null || true
+}
+trap cleanup EXIT
+
+# If the configuration is not loaded and we are in the first terminal
```
Vogtinator, line 67: `jeos-config` shouldn't attempt to read those

files/usr/share/jeos-firstboot/jeos-firstboot-dialogs
Vogtinator, line 58: As you have to touch this anyway, please fix the indentation

files/usr/share/jeos-firstboot/jeos-firstboot-functions
Vogtinator, line 111: This should include a fallback to `en_US` and just echo the result instead of setting a variable.

files/usr/share/jeos-firstboot/jeos-firstboot-functions
```diff
@@ -0,0 +1,136 @@
...
+}
+
+apply_locale()
+{
+	if [ ! -z "$JEOS_LOCALE" ]; then
```
Vogtinator, line 116: The locale should be given as parameter instead

files/usr/share/jeos-firstboot/jeos-firstboot-functions
```diff
@@ -0,0 +1,136 @@
...
+		run langset.sh $JEOS_LOCALE || warn $"Setting the locale failed"
+	fi
+}
+
+apply_locale_and_keytable()
+{
+	if [ ! -z "$JEOS_LOCALE" -a ! -z "$JEOS_KEYTABLE" ]; then
```
Vogtinator, line 123: locale and keytable should be given as parameters

files/usr/share/jeos-firstboot/jeos-firstboot-functions
Vogtinator, line 84: Everything below belongs into `jeos-firstboot-dialogs`, as that's where the code for locale and keymap is.

files/usr/share/jeos-firstboot/jeos-firstboot-dialogs
Vogtinator, line 14: Once `get_current_locale` is fixed, this can just be `default="$(get_current_locale)"`

Vogtinator (commented):
Due to a GitHub accident the review isn't complete yet...

You moved quite some code into functions, so those should use `local` variables when possible now.

files/usr/share/jeos-firstboot/jeos-firstboot-functions
gmoro: This is used just under jeos-config to get the actual current one, should not default inside the function in the logic that is implemented as it's checked before applying in the case of being empty.

files/usr/share/jeos-firstboot/jeos-firstboot-functions
gmoro: This would mean change everything to parameters, now almost all functions are receiving data as global, to ease the work of loading the preset configurations, will put in the list of desirable things in Trello, but will leave like that for now.

files/usr/share/jeos-firstboot/jeos-firstboot-functions
gmoro: It's a generic function to call dialog, the dialog "library" is supposed to have just the actual construction of the dialogs.

Vogtinator:
@gmoro Marking conversations as resolved is ususally up to the reviewer, as otherwise it's not clear whether it's actually resolved.

files/usr/sbin/jeos-config
```diff
@@ -0,0 +1,137 @@
...
+		dialog_network
+		d --infobox $"Restarting network ..." 3 26 || true
+		systemctl restart network
+		;;
+	password)
+		dialog_password
+		apply_password
+		;;
+	*)
+		call_module "$subcommand" "jeos_config"|| echo "Unknown option"
```
Vogtinator, line 136:
```suggestion
		call_module "$subcommand" "jeos_config" || echo "Unknown option"
```

files/usr/share/jeos-firstboot/jeos-firstboot-functions
```diff
@@ -0,0 +1,155 @@
...
+	[ "$(type -t "${module_function}")" = "function" ] || return 1
+	"${module_function}" && true # To not trigger errexit
+	ret=$?
+	[ $ret -eq 0 ] || return $ret
+}
+
+call_module_hook() {
+        local hook="$1"
+        for module in "${modules[@]}"; do
+		call_module ${module} ${hook} || continue
```
Vogtinator, line 43: To restore the same behaviour as before:

```suggestion
		call_module ${module} ${hook} || return $?
```

files/usr/share/jeos-firstboot/jeos-firstboot-functions
```diff
@@ -0,0 +1,155 @@
...
+fi
+
+call_module() {
+	local module="$1"
+	local module_func="$2"
+	module_function="${module}_${module_func}"
+	[ "$(type -t "${module_function}")" = "function" ] || return 1
+	"${module_function}" && true # To not trigger errexit
+	ret=$?
+	[ $ret -eq 0 ] || return $ret
```
Vogtinator, line 37: Just `return $?`

files/usr/share/jeos-firstboot/jeos-firstboot-functions
Vogtinator: If `/etc/locale.conf` doesn't contain `LANG=`, `JEOS_LOCALE` is empty and thus `apply_locale_and_keytable`  doesn't do anything.

files/usr/share/jeos-firstboot/jeos-firstboot-functions
Vogtinator: Currently I view `jeos-firstboot-dialogs` as one file containing multiple modules. That also makes it more obvious what needs to be moved when extracting the "built-in" dialogs as modules. For now it doesn't make any difference which file those functions are defined in, so I don't mind if you want to keep them here.

gmoro:
> @gmoro Marking conversations as resolved is ususally up to the reviewer, as otherwise it's not clear whether it's actually resolved.

Yeah, of course, my bad, I was using it to not confuse myself :P

files/usr/share/jeos-firstboot/jeos-firstboot-functions
Vogtinator: Fortunately it doesn't need many changes, just replace calls to `apply_locale_and_keytable` with `apply_locale_and_keytable "$JEOS_LOCALE" "$JEOS_KEYTABLE"` and it's much clearer, but with unchanged behaviour.

Vogtinator (changes requested):
Much closer!

files/usr/share/jeos-firstboot/jeos-firstboot-functions
gmoro: the "return" will change the behaviour, the current behaviour is actually ignoring errors, for example, the rapsberrywifi does not have a cleanup or post hook, and will fail. I'm thinking in including later one code to guarantee the integrity of the modules itself, check which functions and hooks they really support.

Vogtinator: The current behaviour returns an error when a module hook returns an error. If a hook doesn't exist, that's ignored.
For reference:

```
call_module_hook() {
        local hook="$1"
        for module in "${modules[@]}"; do
                hook_function="${module}_${hook}"
                [ "$(type -t "${hook_function}")" = "function" ] || continue # Ignore if hook doesn't exist
                "${hook_function}" && true # To not trigger errexit
                ret=$?
                [ $ret -eq 0 ] || return $ret # If call failed, return immediately
        done
        return 0
}
```

gmoro: Yes, but I changed because when calling a function that does not exist it should return an error, not ignore it, specially for the jeos-config module calls, so I just left the || continue where it was, ignoring only the hook functions non-existence and actually checking the other functions. It has the exact same effect as before.

files/usr/share/jeos-firstboot/jeos-firstboot-functions
gmoro: I guess so, is just copy and paste from before

files/usr/share/jeos-firstboot/jeos-firstboot-functions
Vogtinator: Time for a `call_module_optional`?

files/usr/share/jeos-firstboot/jeos-firstboot-functions
Vogtinator: Previously it was in a loop, so it was needed to break out of it

files/usr/share/jeos-firstboot/jeos-firstboot-functions
gmoro: Not sure, I spend some time trying to figure the best approach here, and still not sure, should we just guarantee that the modules have all the functions we are calling? What if we add typos by mistake (something I already did during the refactor) and we have no way to really know if the functions are being called or just being ignored because they don't really exist, as a general rule I like to treat all errors, never ignore, so we might need a little bit more code to be sure.

files/usr/share/jeos-firstboot/jeos-firstboot-functions
gmoro: correct

files/usr/share/jeos-firstboot/jeos-firstboot-functions
Vogtinator: Currently `call_module_hook` is used in some critical places, for hooks like `systemd_firstboot` and `post`. Currently this just makes it ignore errors.

I get your argument that `call_module` should fail if the hook isn't implemented, but for `call_module_hook` that's not ideal.

If you don't like `call_module_optional`, `call_module` could accept an optional third parameter which is returned if the hook doesn't exist.

files/usr/share/jeos-firstboot/jeos-firstboot-functions
gmoro: I see the point here, seems default might be necessary indeed as you said.

files/usr/share/jeos-firstboot/jeos-firstboot-functions
gmoro: Just for now, we will probably change this structure again soon enough.

files/usr/share/jeos-firstboot/jeos-firstboot-functions
gmoro: alright, will take a look at this

gmoro: This I will leave to a separate review if you don't mind, this one is already too convoluted and I got politely asked to make smaller changes so it's easier for other people to look at.
As you know I'm taking notes about this smaller details so we can quickly create another PR with this clean-ups.

files/usr/share/jeos-firstboot/jeos-firstboot-dialogs
```diff
@@ -0,0 +1,118 @@
+#!/bin/bash
```
Vogtinator, line 1: Not meant to be executed, so this is misleading

Vogtinator (changes requested):
Reviewed again - no major differences. I should actually give this a test run sometime...

files/usr/share/jeos-firstboot/jeos-firstboot-dialogs
gmoro: Yes, put this one there for some reason I don't remember tho, I think was some problem while testing something, gonna remove.
If you get the chance to test run, that would be cool, but I've been test running at each commit just to be sure, while on that I'm looking a way to make some test automation to have at least a smoke test automated.

Vogtinator: > If you get the chance to test run, that would be cool, but I've been test running at each commit just to be sure,
> while on that I'm looking a way to make some test automation to have at least a smoke test automated.

If you set up a package + image build on OBS, I can do an openQA test run (though the openQA test for jeos-firstboot isn't that extensive currently).

gmoro: Martin already did a openQA run a few commits before, but indeed the tests are not really covering much.

files/usr/sbin/jeos-config
```diff
@@ -0,0 +1,132 @@
...
+trap cleanup EXIT
+
+select_config()
+{
+	modules_order=("locale" '' "keytable" '' "timezone" '' "password" '' "network" '')
+	for module in "${modules[@]}"; do
+		modules_order+=("${module}" '')
+	done
+
+	d --menu $"Select config module" 0 0 $dh_menu "${modules_order[@]}"
```
Vogtinator, line 65:
```suggestion
	d --menu $"Select configuration module" 0 0 $dh_menu "${modules_order[@]}"
```

files/usr/sbin/jeos-config
```diff
@@ -0,0 +1,132 @@
...
+if [ $# -eq 0 ]; then
+	select_config
+	subcommand=${result}
+else
+	subcommand=$1; shift
+fi
+
+case "$subcommand" in
+	locale)
+		dialog_locale
```
Vogtinator, line 109: If there is only a single locale available, this dialog doesn't even appear. At first, I thought `jeos-config` had crashed, so that's something which should be improved.

files/usr/sbin/jeos-config
```diff
@@ -0,0 +1,132 @@
...
+	}
+else
+	run() {
+		"$@"
+	}
+fi
+
+cleanup() {
+	#call_module_hook cleanup
+	echo .oOo.oOo.oOo. > $dialog_out
```
Vogtinator, line 50: I never understood what this is actually for.

files/usr/sbin/jeos-config
```diff
@@ -0,0 +1,132 @@
...
+	rm -f "$dialog_out"
+	# reenable systemd and kernel logs
+	run kill -s SIGRTMAX-10 1
```
Vogtinator, line 53: In `jeos-config`, they're never disabled

files/usr/share/jeos-firstboot/modules/raspberrywifi
```diff
@@ -0,0 +1,152 @@
...
+	while read line; do
+		list+=("${line}" '')
+	done < <(ip link set $wlan_device up && iwlist $wlan_device scan|grep ESSID|cut -d':' -f2|cut -d'"' -f2|sort -u)
+	[ -n "${list[*]}" ]
+}
+
+raspberrywifi_wlan_error()
+{
+	d --title $"Error" --msgbox "$1" 0 0
+	dialog --yesno $"Do you want to retry?" 0 0
```
Vogtinator, line 47: This uses `dialog` instead of `d`, so the "background" is missing. IMO this could just be combined into a single dialog, e.g.

```
Error
---

Your system is on fire.

Retry?

[Yes] [No]
```

files/usr/sbin/jeos-config
```diff
@@ -0,0 +1,132 @@
...
+	run setterm -msg on 2>/dev/null || true
+}
```
Vogtinator, line 55: Can you just add an `echo` here, so that there's a new blank line after exit? That way it looks much better. Same for `jeos-firstboot`.

files/usr/sbin/jeos-config
```diff
@@ -0,0 +1,132 @@
...
+		dialog_network
+		d --infobox $"Restarting network ..." 3 26 || true
+		systemctl restart network
+		;;
+	password)
+		dialog_password
+		apply_password
+		;;
+	*)
+		call_module "$subcommand" "jeos_config" || echo "Unknown option"
```
Vogtinator, line 131:
```suggestion
		call_module "$subcommand" "jeos_config" || echo "Unknown option '$subcommand'"
```

files/usr/sbin/jeos-config
```diff
@@ -0,0 +1,132 @@
...
+			;;
+		\?) 
+			echo "Invalid Option: -$OPTARG" 1>&2
+			usage
+     			exit 1
+			;;
+	esac
+done
+
+if [ $# -eq 0 ]; then
```
Vogtinator, line 100: This should use `${OPTIND}` set by `getopts` to avoid issues like

```
e111:~ # jeos-config -- locale
/usr/share/jeos-firstboot/jeos-firstboot-functions: line 34: type: --: invalid option
type: usage: type [-afptP] name [name ...]
Unknown option
```

files/usr/sbin/jeos-config
```diff
@@ -0,0 +1,132 @@
...
+cat <<EOF
+Usage: jeos-config [OPTION...] [CONFIG_NAME]
+Configure system settings using an interactive dialog
+
+	-h		shows this usage help
+	locale		Show configuration for locale
+	keytable	Show configuration for keyboard
+	timezone	Show configuration for timezone
+	password	Show configuration for password
+	network		Show configuration for network
```
Vogtinator, line 79: The `network` module doesn't show or offer any configuration - it only sets up DHCP. Maybe the module should be renamed as well?

Vogtinator: Actually, this module not only overwrites network configuration, it also forces a new DHCP client identifier, which means that the system gets a new IP each time this is called. This has to be fixed.

Vogtinator (changes requested):
Did some testing, found some issues.

files/usr/sbin/jeos-config
gmoro: I think is to avoid leaking any info, erasing any response from the file if anything goes wrong and it does not get deleted.

gmoro: But is just a guess, will put in the list of thinks to check

Vogtinator: Probably so that it doesn't end up on the hard drive or something like that.

files/usr/sbin/jeos-config
gmoro: Theoretically it is a reconfiguration, so redoing the whole config from scratch is what is supposed to do.
For now seems useless, but I will implement extra checks later on, specially for multiple interfaces and also for disconnected cables, that should make this module useful as it is now.

Vogtinator: It means that you absolutely must not call this over SSH or such, so at the very least there has to be a warning!

gmoro: Added a warning

gmoro:
Anything else missing, would be good to merge that soon as there are other stuff to be merged.

files/usr/share/jeos-firstboot/jeos-firstboot-functions
```diff
@@ -0,0 +1,156 @@
...
+                [ -d "/usr/lib/locale/${locale}.utf8" ] || continue
+                list+=("${locale}" '')
+        done
+        [ -n "$list" ]
+}
+
+get_current_locale()
+{
+	cur_locale=`awk -F= '$1 == "LANG" { print $2; exit }' /etc/locale.conf`
+	[ -n "$cur_locale" ] && cur_locale="en_US"
```
Vogtinator, line 130: This is the wrong way around, it either has to be

```
[ -z "$cur_locale" ] && cur_locale="en_US"
```

or

```
[ -n "$cur_locale" ] || cur_locale="en_US"
```

files/usr/sbin/jeos-config
```diff
@@ -0,0 +1,142 @@
...
+		"$@"
+	}
+fi
+
+cleanup() {
+	#call_module_hook cleanup
+	echo .oOo.oOo.oOo. > $dialog_out
+	rm -f "$dialog_out"
+	# reenable systemd and kernel logs
+	run setterm -msg on 2>/dev/null || true
```
Vogtinator, line 53: This also isn't needed here

files/usr/share/jeos-firstboot/jeos-firstboot-dialogs
Vogtinator: Shebang is still there

Vogtinator (changes requested):
Found a bug, will do a test run again

files/usr/share/jeos-firstboot/jeos-firstboot-functions
```diff
@@ -0,0 +1,156 @@
...
+                fi
+        done
+        popd &>/dev/null
+fi
+
+call_module() {
+	local module="$1"
+	local module_func="$2"
+	module_function="${module}_${module_func}"
+	[ "$(type -t "${module_function}")" = "function" ] || return 1
```
Vogtinator, line 33:
```suggestion
	[ "$(type -t -- "${module_function}")" = "function" ] || return 1
```

To avoid

```
f32:~ # /usr/sbin/jeos-config -- --help
/usr/share/jeos-firstboot/jeos-firstboot-functions: line 33: type: --: invalid option
type: usage: type [-afptP] name [name ...]
Unknown option '--help'
```

Vogtinator (changes requested):
Gave it a test run

files/usr/share/jeos-firstboot/modules/raspberrywifi
```diff
@@ -0,0 +1,151 @@
...
+	local line
+	while read line; do
+		list+=("${line}" '')
+	done < <(ip link set $wlan_device up && iwlist $wlan_device scan|grep ESSID|cut -d':' -f2|cut -d'"' -f2|sort -u)
+	[ -n "${list[*]}" ]
+}
+
+raspberrywifi_wlan_error()
+{
+	dialog --title $"Error" --yesno $"$1\n\nDo you want to retry?" 0 0
```
Vogtinator, line 46: For the proper background:

```suggestion
	d --title $"Error" --yesno $"$1\n\nDo you want to retry?" 0 0
```

Vogtinator (changes requested):
There are some more instances of `dialog` which should be `d` instead, but that's something for later.

files/usr/share/jeos-firstboot/modules/raspberrywifi
gmoro: I will check that, but I think I used dialog to avoid the logic inside the d function,that was causing some inappropriate UX, if I remember correctly

files/usr/sbin/jeos-config
```diff
@@ -0,0 +1,141 @@
...
+		dialog_keytable
+		JEOS_LOCALE="$(get_current_locale)"
+		apply_locale_and_keytable
+		;;	
+	timezone)
+		dialog_timezone
+		timedatectl set-timezone "$JEOS_TIMEZONE"
+		;;	
+	network)
+		if ! d --yesno $"Network will be restarted\nDo you want to continue?" 6 30; then
```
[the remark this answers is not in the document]
gmoro: What is your suggestion? Maybe change it to "network will be reconfigured and restarted", and why does it changed the client ID?

files/usr/share/jeos-firstboot/jeos-firstboot-functions
gmoro: so still not enough the OPTIND change?

files/usr/share/jeos-firstboot/modules/raspberrywifi
Vogtinator: Indeed - the current design of `d` doesn't allow `--yesno` as it only allows a zero exit value.

files/usr/share/jeos-firstboot/jeos-firstboot-functions
Vogtinator: The OPTINID change is fine, now it properly tries to execute `--help` as module. That's not a valid module name though, so it has to be treated specially until it's validated by `type`.

files/usr/sbin/jeos-config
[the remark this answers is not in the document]
Vogtinator: > What is your suggestion? Maybe change it to "network will be reconfigured and restarted",

Sounds good, maybe "This will create a new network configuration from scratch, all connections will be lost."?

> and why does it changed the client ID?

Absolutely no idea.

files/usr/sbin/jeos-config
```diff
@@ -125,7 +125,8 @@ case "$subcommand" in
 		timedatectl set-timezone "$JEOS_TIMEZONE"
 		;;	
 	network)
-		if ! d --yesno $"Network will be restarted\nDo you want to continue?" 6 30; then
+		if ! d --yesno $"This will create a new network configuration from scratch, \
```
Vogtinator, line 128: Is that `\` necessary?

gmoro: Not really

Vogtinator (approved):

PR #75, opened 2021-01-05
Use race-free DBus API to silence systemd

Vogtinator:
Only available in 246+, so keep a fallback.

Not really tested yet.


gmoro (approved):
Tested this before, works as expected, going to merge

PR #78, opened 2021-03-11
Improve menu appearance

Vogtinator:
Consistent appearance and less empty space inside the dialog.

While `dialog --menu` allows to specify the total dialog width, height and menu item count, I was unable to find a way to get the dialog box just small enough to contain the description and menu items. There are always two empty lines below the menu list.

Everything automatic results in empty lines and will overflow the backtitle if there are too many items:

 `dialog --menu "Choose wisely" 0 0 0 Value1 "" Value2 ""`:
Screenshot_20210311_163552

The overflowing can be avoided by setting the menu height explicitly, which is what this PR does, but the empty lines remain:

 `dialog --menu "Choose wisely" 0 0 2 Value1 "" Value2 ""`:
Screenshot_20210311_163552

By specifying the dialog height as items+5, the empty lines disappear:

`dialog --menu "Choose wisely" 7 0 0 Value1 "" Value2 ""`
Screenshot_20210311_163848

For some reason the dialog height is ignored when there are more items than can fit - it fills the entire screen anyway!
`dialog --menu "Choose wisely" 20 0 0 $(seq 1 100)`
(screenshot omitted)

However, by specifying the menu item height explicitly as well, it's suddenly honoured again, even if the menu height is too big to fit:
`dialog --menu "Choose wisely" 20 0 30 $(seq 1 100)`
Screenshot_20210311_164410

The behaviour gets really weird in that case though, because
`dialog --menu "Choose wisely" 7 0 0 Value1 "" Value2 ""` and `dialog --menu "Choose wisely" 7 0 2 Value1 "" Value2 ""` behave differently, even though they shouldn't. I assume this is because of the two empty lines at the bottom.

Any idea?

Marked as draft because I didn't check it inside jeos-firstboot yet, only outside.

### DIFF
--- a/files/usr/share/jeos-firstboot/jeos-firstboot-dialogs
+++ b/files/usr/share/jeos-firstboot/jeos-firstboot-dialogs
@@ -21,7 +21,7 @@ dialog_locale()
         elif [ "${#list[@]}" -eq 2 ]; then # Only a single entry
                 newlocale="${list[0]}"
         else
-                d --default-item "$default" --menu $"Select system locale" 0 0 $dh_menu "${list[@]}"
+                d --default-item "$default" --menu $"Select system locale" 0 0 "$(menuheight ${#list[@]})" "${list[@]}"
                 newlocale="${result}"
         fi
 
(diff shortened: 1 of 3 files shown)


Vogtinator:
Apparently this bug got fixed in dialog recently!

https://openqa.opensuse.org/tests/1692837#step/firstrun/5

So we just need https://bugzilla.opensuse.org/show_bug.cgi?id=1184503 fixed and the `--backtitle` additions here

PR #79, opened 2021-04-21
Improve dialogs, support manually entering SSID for WiFi

gmoro:
Introduce wifi hidden SSID supports
Improve menu appearance fixing bsc#1184503

### DIFF
--- a/files/usr/share/jeos-firstboot/modules/raspberrywifi
+++ b/files/usr/share/jeos-firstboot/modules/raspberrywifi
@@ -89,9 +89,8 @@ do
         break
     fi
 	if [ "$wlan_network" == "manual" ]; then
-		if d --inputbox $"SSID of hidden network" 0 0; then
-			wlan_network="$result"
-		fi
+		d --inputbox $"SSID of hidden network" 0 0
+		wlan_network="$result"
 	else
 		wlan_network="$(cut -d "=" -f2- <<< $wlan_network)"
 	fi
--- a/files/usr/share/jeos-firstboot/modules/raspberrywifi
+++ b/files/usr/share/jeos-firstboot/modules/raspberrywifi
@@ -75,7 +75,8 @@ do
         wlan_device="${result}"
     fi
 
-    if raspberrywifi_get_wlan_networks && d --no-tags --menu  $"Select wireless network to connect" 0 0 "$(menuheight ${#list[@]})" "${list[@]}"; then
+    if raspberrywifi_get_wlan_networks; then
+        d --no-tags --menu $"Select wireless network to connect" 0 0 "$(menuheight ${#list[@]})" "${list[@]}"
         wlan_network="$result"
     else
         if raspberrywifi_wlan_error $"Error listing wireless networks"; then
--- a/files/usr/share/jeos-firstboot/modules/raspberrywifi
+++ b/files/usr/share/jeos-firstboot/modules/raspberrywifi
@@ -36,7 +36,11 @@ raspberrywifi_get_wlan_networks()
 	list=()
 	local line
 	while read line; do
+	if [ -z $line ]; then
+                list+=("()" '')
+        else
 		list+=("${line}" '')
+	fi
 	done < <(ip link set $wlan_device up && iwlist $wlan_device scan|grep ESSID|cut -d':' -f2|cut -d'"' -f2|sort -u)
 	[ -n "${list[*]}" ]
 }
--- a/files/usr/share/jeos-firstboot/modules/raspberrywifi
+++ b/files/usr/share/jeos-firstboot/modules/raspberrywifi
@@ -41,7 +41,7 @@ raspberrywifi_get_wlan_networks()
         else
 		list+=("${line}" '')
 	fi
-	done < <(ip link set $wlan_device up && iwlist $wlan_device scan|grep ESSID|cut -d':' -f2|cut -d'"' -f2|sort -u)
+	done < <(ip link set $wlan_device up && iwlist $wlan_device scan|grep ESSID|cut -d':' -f2|cut -d'"' -f2)
 	[ -n "${list[*]}" ]
 }
 
--- a/files/usr/share/jeos-firstboot/modules/raspberrywifi
+++ b/files/usr/share/jeos-firstboot/modules/raspberrywifi
@@ -84,6 +84,11 @@ do
         fi
         break
     fi
+    if [ "$wlan_network" == "()" ]; then
+	if d --inputbox $"SSID of hidden network" 0 0; then
+		wlan_network="$result"
+	fi
+    fi
 
     d --menu $"Select authentication mode" 0 0 $dh_menu $"WPA-PSK" '' $"WPA-EAP" '' $"Open" ''
     wlan_auth_mode="$result"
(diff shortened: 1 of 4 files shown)


Vogtinator (changes requested):
Two of those commits are from me, please fix the authorship.

The approach with `()` has some issues:
- There can be multiple hidden networks
- `()` is a valid ESSID
- Hidden networks don't broadcast, so they might not show up at all

So instead I suggest to always add a special entry to enter a SSID manually. `dialogs` supports keys and values in dialogs, so you can use that to distinguish between the ESSIDs and the special option.

gmoro:
> Two of those commits are from me, please fix the authorship.
I just copied and paste this, do you really need the authorship?
> 
> The approach with `()` has some issues:
> 
>     * There can be multiple hidden networks
Forgot to implement that
> 
>     * `()` is a valid ESSID
Thats the main problem
> 
>     * Hidden networks don't broadcast, so they might not show up at all
Hidden networks don't broadcast the SSID but they are always visible for iwlist, with an empty SSID
I thought that was not a good enough solution, but now we have something to go from.
> 
> 
> So instead I suggest to always add a special entry to enter a SSID manually. `dialogs` supports keys and values in dialogs, so you can use that to distinguish between the ESSIDs and the special option.
I did try to do that initially, but failed, do you have a good example of doing this special entries?

Vogtinator:
> > Two of those commits are from me, please fix the authorship.
> I just copied and paste this,

Copy + paste the entire diff and commit message separately? If you use `git cherry-pick` or just work on the branch, it'll keep them with proper author and date.

>  do you really need the authorship?

Yes please.

> > The approach with `()` has some issues:
> > ```
> > * There can be multiple hidden networks
> > ```
> 
> Forgot to implement that
> 
> > ```
> > * `()` is a valid ESSID
> > ```
> 
> Thats the main problem
> 
> > ```
> > * Hidden networks don't broadcast, so they might not show up at all
> > ```
> 
> Hidden networks don't broadcast the SSID but they are always visible for iwlist, with an empty SSID

I read up on that and indeed, they still send beacons regularly.

> I thought that was not a good enough solution, but now we have something to go from.
>
> > So instead I suggest to always add a special entry to enter a SSID manually. `dialogs` supports keys and values in dialogs, so you can use that to distinguish between the ESSIDs and the special option.
> 
> I did try to do that initially, but failed, do you have a good example of doing this special entries?

`dialog --no-tags --menu "Choose wisely" 0 0 0 "SSID=SSID 1" "SSID 1" "SSID=SSID 2" "SSID 2" "manual" "Enter SSID manually"`

gmoro:
By the way, I need to retab the whole thing, it's a mess of mixed tabs and spaces, I guess should wait until the next release, what do you think?

files/usr/share/jeos-firstboot/modules/raspberrywifi
```diff
@@ -35,15 +35,23 @@ raspberrywifi_get_wlan_networks()
 {
 	list=()
 	local line
+	local has_hidden=false
```
Vogtinator, line 38: Looks like booleans, but isn't really. To avoid this trap:

```
local has_hidden=
...
has_hidden=1
...
if [[ ${has_hidden} ]]; then ...
fi
```

Vogtinator (changes requested):
This has to make use of the key/value split in menu dialogs. Otherwise you'll always get conflicts between SSIDs and whatever the option for the hidden network is called internally.

Vogtinator:
> By the way, I need to retab the whole thing, it's a mess of mixed tabs and spaces, I guess should wait until the next release, what do you think?

IMO that can be done at any point, but in a separate PR. It can be mentioned in the `.git-blame-ignore-revs` file

files/usr/share/jeos-firstboot/modules/raspberrywifi
```diff
@@ -68,20 +76,27 @@ do
...
         wlan_network="$result"
     else
         if raspberrywifi_wlan_error $"Error listing wireless networks"; then
             continue
         fi
         break
     fi
-
-    d --menu $"Select authentication mode" 0 0 $dh_menu $"WPA-PSK" '' $"WPA-EAP" '' $"Open" ''
+	if [ "$wlan_network" == "manual" ]; then
```
Vogtinator, line 91: Broken indentation

files/usr/share/jeos-firstboot/modules/raspberrywifi
```diff
@@ -68,20 +76,27 @@ do
...
+		if d --inputbox $"SSID of hidden network" 0 0; then
```
Vogtinator, line 92: `d` never fails

files/usr/share/jeos-firstboot/modules/raspberrywifi
gmoro: tabs vs spaces again, I will fix this all at once, changing everything to tabs in the new code.

files/usr/share/jeos-firstboot/modules/raspberrywifi
gmoro: forgot about that

files/usr/share/jeos-firstboot/modules/raspberrywifi
```diff
@@ -35,15 +35,23 @@ raspberrywifi_get_wlan_networks()
 {
 	list=()
 	local line
+	local has_hidden=
 	while read line; do
-		list+=("${line}" '')
-	done < <(ip link set $wlan_device up && iwlist $wlan_device scan|grep ESSID|cut -d':' -f2|cut -d'"' -f2|sort -u)
+	if [ -z $line ]; then
```
Vogtinator, line 40: IMO the `Enter SSID manually` entry could always be shown

gmoro: I guess it's better, because now testing it again, sometimes the iwlist misses the scan, so I will do that instead

Vogtinator (commented):
LGTM code wise.

Now `raspberrywifi_get_wlan_networks` never fails, because even when `ip link` or `iwlist` fail, it has the `manual` entry which makes `$list` non-empty. Maybe that should be caught explicitly? Though I'm not sure how

gmoro:
> LGTM code wise.
> 
> Now `raspberrywifi_get_wlan_networks` never fails, because even when `ip link` or `iwlist` fail, it has the `manual` entry which makes `$list` non-empty. Maybe that should be caught explicitly? Though I'm not sure how

I guess that makes the previous behaviour wrong then, because an empty list is a valid thing, there might be no networks in range, we should really start checking return codes for the commands, only way to guarantee

files/usr/share/jeos-firstboot/modules/raspberrywifi
```diff
@@ -68,20 +71,26 @@ do
...
         wlan_device="${list[0]}"
     else
-        d --menu  $"Select wireless card to configure" 0 0 $dh_menu "${list[@]}"
+        d --menu  $"Select wireless card to configure" 0 0 "$(menuheight ${#list[@]})" "${list[@]}"
         wlan_device="${result}"
     fi
 
-    if raspberrywifi_get_wlan_networks && d --menu  $"Select wireless network to connect" 0 0 $dh_menu "${list[@]}"; then
+    if raspberrywifi_get_wlan_networks && d --no-tags --menu  $"Select wireless network to connect" 0 0 "$(menuheight ${#list[@]})" "${list[@]}"; then
         wlan_network="$result"
```
Vogtinator, line 79:
```suggestion
    if raspberrywifi_get_wlan_networks; then
        d --no-tags --menu $"Select wireless network to connect" 0 0 "$(menuheight ${#list[@]})" "${list[@]}"
        wlan_network="$result"
```

Vogtinator (approved):

PR #82, opened 2021-05-11
Introduce welcome screen for console switching

Vogtinator:
Currently, jeos-firstboot is only on the active console, which can only be
configured by changing the kernel cmdline. This commit introduces a welcome
dialog which is shown on all active consoles (from /proc/consoles) and it
continues on the console where the welcome screen was accepted.

Screenshot_20210512_102352

~WIP because:~

- [x] Welcome screen needs proper text
- [x] When started from systemd, it gets `/dev/console` as tty. That's not handled properly yet.
- [ ] If there's just a single entry in `/proc/consoles`, skip the screen?


Vogtinator:
Test on hardware was a success: https://bugzilla.suse.com/show_bug.cgi?id=1184157#c9

gmoro (approved):
Looks good to me

PR #83, opened 2021-07-08
Skip the lo interface when probing for DHCP

Vogtinator:
It's unnecessary and causes wicked to write an error into the journal.


gmoro (approved):
LGTM, good work!

PR #84, opened 2021-11-22
Use the snapper config to determine whether to run setup-quota (boo#1192940)

Vogtinator:
It's possible that btrfs quotas are enabled, but the snapper-specific
configuration wasn't done yet. Check for the QGROUP snapper config directly
and call snapper setup-quota if not set. It can deal with already enabled
btrfs quotas as well.

### DIFF
--- a/files/usr/sbin/jeos-firstboot
+++ b/files/usr/sbin/jeos-firstboot
@@ -228,8 +228,9 @@ fi
 
 # Test if snapper is available
 if [ -x /usr/bin/snapper -a "$(stat --format=%T -f /)" = "btrfs" ]; then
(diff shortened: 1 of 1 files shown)


Vogtinator:
Also, meanwhile the `jeos-firstboot` service is run when DBus is up, so we could possibly drop the `--no-dbus`. That avoids issues if `snapperd.service` is already started in the background and thus might miss the config change. On the other hand, it triggers `snapperd.service` to start, so might result in a deadlock if the dependencies are changed in an incompatible way.

Vogtinator:
Potentially we could also remove this entirely, and rely completely on Kiwi to set this up.

Vogtinator:
Closing in favor of #86.

PR #86, opened 2021-11-29
Drop call to snapper setup-quota, kiwi does that meanwhile

Vogtinator:
Alternative to https://github.com/openSUSE/jeos-firstboot/pull/84.

Intentionally not mentioning boo#1192940 because this just removes already unused code...


gmoro (approved):
Looks ok

PR #87, opened 2022-01-31
Some fixes for jeos-firstboot-snapshot

Vogtinator:
Make it ready for MicroOS and fix two other minor issues.


lnussel:
would ConditionPathIsMountPoint in the service file do the job inside the service file instead of the shell script?

Vogtinator:
> would ConditionPathIsMountPoint in the service file do the job inside the service file instead of the shell script?

Then we'd need yet another unit just to remove `/var/lib/YaST/reconfig_system`

lnussel:
Indeed. OTOH that would make it more clear as removing the file wouldn't be a side effect. anyway, go ahead :)

gmoro (approved):
Looks good to me

PR #88, opened 2022-02-15
Skip broken consoles (bsc#1195975)

Vogtinator:
Currently it starts dialog for all consoles listed in /proc/consoles, but
some of them might not be working (e.g. not set up, see also bsc#1175514#c26).
When dialog is used on one of those consoles, it attempts to use stderr first
and if even that fails, /dev/tty. As a result, starting dialog on a broken
console causes a conflict with dialog on a working console, and weird behaviour
occurs.

Use "stty size" to check whether the console is usable before starting dialog.


gmoro (approved):
Looks good

Vogtinator:
I added another somewhat related tiny fix on top

PR #94, opened 2022-04-11
Fix size of the "No root password set" dialog

Vogtinator:
The warn function hardcoded the size and the values were too small for this
message, so the important part was cut off and only visible through scrolling.
Fix that by always using automatic size.

The other warning dialogs look a bit odd now (not quite wide enough), but it's readable and they should never show up anyway...


gmoro (approved):
The whole error handling needs to be rechecked, for now this will make at least sure everything is readable

PR #96, opened 2022-04-13
Make use of SPDX identifiers

### LINKED ISSUE #93: Use SPDX identifiers instead of license headers
We can just use
```
# SPDX-License-Identifier: MIT
# SPDX-FileCopyrightText: 2016-2022 SUSE LLC (or whatever it's called now...)
```

instead of the large license header in each file.

See https://spdx.dev/ids/#how.

Vogtinator:
Replace license headers and copyright strings with SPDX lines.

Fixes #93

Untested.

### DIFF
--- a/files/usr/lib/systemd/system/jeos-firstboot.service
+++ b/files/usr/lib/systemd/system/jeos-firstboot.service
@@ -1,9 +1,5 @@
-#  This file is part of systemd.
-#
-#  systemd is free software; you can redistribute it and/or modify it
-#  under the terms of the GNU Lesser General Public License as published by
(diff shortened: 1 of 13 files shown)


files/usr/lib/systemd/system/jeos-firstboot.service
```diff
@@ -1,9 +1,5 @@
-#  This file is part of systemd.
-#
-#  systemd is free software; you can redistribute it and/or modify it
-#  under the terms of the GNU Lesser General Public License as published by
```
Vogtinator, line 4: This was actually never from systemd...

gmoro (approved):
Good!

Issue #102, opened 2022-11-29
jeos-firstboot reads only /etc/os-release


thkukuk:
The jeos-firstboot script reads only /etc/os-release at the beginning of that file, but does not fallback to /usr/lib/os-release

Vogtinator:
https://github.com/openSUSE/jeos-firstboot/pull/103, untested

Vogtinator:
#103 was merged

PR #103, opened 2022-11-29
Support /usr/lib/os-release (#102)

Vogtinator:
Move reading os-release to jeos-firstboot-function and use /usr/lib/os-release if /etc/os-release is not available.


thkukuk:
Tested and works

PR #105, opened 2023-06-06
remove --setup-machine-id from systemd-firstboot

lnussel:
No longer works in running system, see
https://github.com/systemd/systemd/pull/27750


lnussel:
look like this one got lost

PR #97, opened 2022-04-14
Add support for shellcheck

Vogtinator:
TODO:

- [ ] Apply manual fixups
- [ ] Test
- [ ] Integrate into CI

Fixes #95

PR #107, opened 2023-09-27
Order modules, read from /etc 

lnussel:


files/usr/share/jeos-firstboot/jeos-firstboot-functions
```diff
@@ -54,6 +66,7 @@ call_module() {
 
 call_module_hook() {
         local hook="$1"
+	_init_modules
```
Vogtinator, line 69: Should be called in global scope on inclusion, like before.

files/usr/share/jeos-firstboot/jeos-firstboot-functions
```diff
@@ -26,14 +26,26 @@ list=
...
+			fi
+		done
+	done
+}
+
+_init_modules() {
+	[ -z "$modules" ] || return 0
+	local module f
+	while read -r module f; do
+		[ -L "$f" ] && [ "$(readlink "$f")" = "/dev/null" ] && continue
```
Vogtinator, line 44: IMO this needs to be a full if. The `&&` shorthand syntax should only be used once per line.

lnussel: :shrug:

files/usr/share/jeos-firstboot/jeos-firstboot-functions
lnussel: moved it to after dryrun etc in main file. doing that on source seems ugly to me

lnussel:
ping?

files/usr/lib/systemd/system/jeos-firstboot.service
```diff
@@ -34,6 +34,7 @@ ExecStart=/usr/sbin/jeos-firstboot
 StandardOutput=tty
 StandardInput=tty
 #StandardError=tty
+KeyringMode=shared
```
Vogtinator, line 37: Needs an explanation.

files/usr/sbin/jeos-firstboot
```diff
@@ -4,14 +4,22 @@
 
 set -e
 
+# set to --keep-tite to use altenate screen. Better for debugging but causes flicker
```
Vogtinator, line 7:
```suggestion
# set to --keep-title to use alternate screen. Better for debugging but causes flicker
```

files/usr/sbin/jeos-firstboot
```diff
@@ -4,14 +4,22 @@
...
+dialog_altenate_screen=
```
Vogtinator, line 8: Typo. This is used in jeos-firstboot-functions and needs to be defined there. Currently it's undefined when running jeos-config.

Vogtinator:
I'll rebase #97. Can you then rebase this on that PR instead?

Vogtinator (changes requested):

lnussel:
if you add 97 that would screw up mine I guess? Would be unfair after a year letting that other thing linger

files/usr/share/jeos-firstboot/jeos-firstboot-functions
```diff
@@ -110,11 +124,18 @@ findlocales()
         list=()
         local l locale
         # List only locales which are both in live-langset-data and glibc-locale(-base)
-        for l in /usr/share/langset/*; do
-                locale="${l#/usr/share/langset/}"
-                [ -d "/usr/lib/locale/${locale}.utf8" ] || continue
-                list+=("${locale}" '')
-        done
+        if [ -d /usr/share/langset ]; then
```
Vogtinator, line 127: jeos-firstboot calls langset.sh so this is a hard requirement

Vogtinator:
> if you add 97 that would screw up mine I guess?

I thought I wrote in here that this duplicates some parts of #97, but I can't find that comment. Maybe I never submitted it...

>  Would be unfair after a year letting that other thing linger

IMO this PR is a reason to finish up #97, while previously it was just "nice to have".

files/usr/sbin/jeos-firstboot
lnussel: it's actually called --keep-tite

files/usr/share/jeos-firstboot/jeos-firstboot-functions
lnussel: yeah, was good enough for dry run though

lnussel:
well, you can rebase on this one if you have to rebase anyway :)

Vogtinator:
> well, you can rebase on this one if you have to rebase anyway :)

"Fix calling without installation" is incomplete and I personally don't like `type -P $0` or using the `${pfx%/*/*}` syntax. At that point it doesn't matter much whether #97 is changed to be more like the commit here or the other way around. I'll just merge them both together and add a `Co-authored-by` tag. Does that sound good?

lnussel:
fine with me. calling dirname etc requires external programs for something bash can do by itself. bad style :)

lnussel:
rebased on yours

lnussel:
ok now?

files/usr/share/jeos-firstboot/jeos-firstboot-functions
```diff
@@ -27,15 +27,28 @@ list=
...
+}
+
+init_modules() {
+	[ -z "$modules" ] || return 0
+	local module f
+	while read -r module f; do
+		if [ -L "$f" ] && [ "$(readlink "$f")" = "/dev/null" ]; then
+			continue
+		fi
+		. "$f"
```
Vogtinator, line 48: Needs to do `source "$f"` within an if to not kill the entire process if a module has a syntax error

lnussel:
updated

Vogtinator (approved):

PR #109, opened 2023-11-28
Add KeyringMode=shared

lnussel:
In case the initrd puts keys in the kernel keyring, e.g. for disk encryption the firstboot wizard may want to access them.


Vogtinator:
This is part of https://github.com/openSUSE/jeos-firstboot/pull/107 already?

lnussel:
yes but you are not merging that one. We need this fix though

PR #111, opened 2024-01-17
Add status_mail module for systemd-status-mail configuration

Vogtinator:
Screenshot_20240117_115527

Depends on #97 

Draft because systemd-status-mail does not support SMTP auth yet.

### DIFF
--- a/files/usr/share/jeos-firstboot/modules/status_mail
+++ b/files/usr/share/jeos-firstboot/modules/status_mail
@@ -1,0 +1,97 @@
+# SPDX-License-Identifier: MIT
+# SPDX-FileCopyrightText: Copyright 2024 SUSE LLC
+# shellcheck shell=bash
+
+# This also gets the password as value, so make sure not to leak that in the process list
+status_mail_set_config_option()
+{
+	key="$1"
+	value="$2"
+
+	# Create the config file in /etc if necessary
+	[ -e /etc/default/systemd-status-mail ] || touch /etc/default/systemd-status-mail
+
+	# Ensure the permissions are correct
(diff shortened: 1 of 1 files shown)


Vogtinator:
> What about including the module in the systemd-status-mail package directly?

I thought about that as well. The "API" for the config file is more stable than the jeos-firstboot module API, other than that there's no reason. Ultimately it won't matter much.

> Also IMO the module should not be called on first boot but only via jeos-config. The firstboot wizard is meant to query mandatory stuff rather than optional convenience features

Agreed, just needs to be documented then for visibility.

files/usr/share/jeos-firstboot/modules/status_mail
```diff
@@ -0,0 +1,95 @@
...
+	d --title $"E-Mail Notifications" \
+		--insecure \
+		--mixedform $"If desired, enter a mail address to receive system status notifications:" 8 0 8 \
+		$"Destination" 1 0 "$ADDRESS" 1 18 35 0 0 \
+		$"Sender address" 2 0 "$FROM" 2 18 35 0 0 \
+		$"SMTP configuration (optional):" 4 0 "" 4 35 0 0 2 \
+		$"Server" 5 0 "$RELAY_HOST" 5 18 35 0 0 \
+		$"Authentication (if needed):" 6 0 "" 6 35 0 0 2 \
+		$"Username" 7 0 "$MAILX_AUTH_USER" 7 18 35 0 0 \
+		$"Password" 8 0 "$MAILX_AUTH_PASSWORD" 8 18 35 0 1
```
Vogtinator, line 57: Needs special handling actually

laenion (approved):

PR #112, opened 2024-01-18
Add ssh_enroll module

Vogtinator:
Screenshot_20240118_165514

Screenshot_20240118_165535

TODO:

* Depends on #113
* ~There's no cleanup if `d` exits~
* ~Deselecting all keys does not fully work~

### DIFF
--- a/files/usr/share/jeos-firstboot/modules/ssh_enroll
+++ b/files/usr/share/jeos-firstboot/modules/ssh_enroll
@@ -12,11 +12,9 @@ ssh_enroll_handle_pubkeys()
 
 	items=()
 	lineno=1
-	state=on
 	while read -r type key comment; do
-		items+=("$lineno" "${type} ${key:0:8}...${key: -20:20}" "$state")
+		items+=("$lineno" "${type} ${key:0:8}...${key: -20:20}" "off")
 		lineno=$((lineno+1))
-		state=off # Only import the first key by default
 	done <"$filename"
 
 	text="$(printf $"Received %d public keys from %s. Select the key(s) to import." "$(wc -l <"$filename")" "$(awk '{print $3; exit}' "$filename")")"
--- a/files/usr/share/jeos-firstboot/modules/ssh_enroll
+++ b/files/usr/share/jeos-firstboot/modules/ssh_enroll
@@ -77,11 +77,23 @@ ssh_enroll_do_config()
 		echo
 		echo $"Please verify that the host key matches:"
 		for i in /etc/ssh/ssh_host_*key.pub; do ssh-keygen -l -f "$i"; done | awk '{ print $2" "$4 }'
+		echo
+
+		echo $"Randomart for ssh -o \"VisualHostKey yes\":"
+		# Display the randomart for all key types next to each other
+                allartlines=()
+                for i in /etc/ssh/ssh_host_*key.pub; do
+                        readarray -t artlines < <(ssh-keygen -l -v -f "$i" | tail -n+2)
+                        for (( j=0; j < "${#artlines[@]}"; j++ )); do
+                                allartlines[$j]="${allartlines[$j]}${artlines[$j]} "
+                        done
+                done
+                ( IFS=$'\n'; echo -n "${allartlines[*]}" )
 	} >"${tmpdir}/msg"
 
 	(
 		# Have to use dialog directly here, no subprocesses allowed
-		dialog --backtitle "$PRETTY_NAME" --title $"SSH Key Enrollment" --exit-label $"Cancel" --textbox "${tmpdir}/msg" 12 70 &
+		dialog --backtitle "$PRETTY_NAME" --title $"SSH Key Enrollment" --exit-label $"Cancel" --textbox "${tmpdir}/msg" 0 0 &
 		pid=$!; trap 'kill $pid' SIGTERM; echo dialog >"${tmpdir}/dialogfifo"; wait $pid; echo dialog $? >>"${tmpdir}/exitfifo"
 	) &
 	dialogpid=$!
--- a/files/usr/share/jeos-firstboot/modules/ssh_enroll
+++ b/files/usr/share/jeos-firstboot/modules/ssh_enroll
@@ -56,21 +56,7 @@ ssh_enroll_do_config()
 	tmpdir="$(mktemp -d)"
 
 	# Start the server in the background and show a dialog with information while it's running
-	# and offer the option to cancel the process. Unfortunately bash doesn't have a way to wait for
-	# one of multiple processes to exit and get information about which process exited and how.
-	# To work around that, use subprocesses with FIFOs: One to signal once they're up and can handle
-	# SIGTERM properly and one to signal that it exited and how.
-	# See also the welcome-screen implementation.
-	mkfifo "${tmpdir}/serverfifo"
-	mkfifo "${tmpdir}/dialogfifo"
-	mkfifo "${tmpdir}/exitfifo"
-
-	(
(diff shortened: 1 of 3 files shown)


lnussel:
this is pretty cool, maybe we can have it standalone?

Vogtinator:
> this is pretty cool, maybe we can have it standalone?

Thought about that as well. It would either have to be integratable into jeos-firstboot to avoid double maintenance or be small enough to have neglegible overhead.

I'd also like to replace the background process + fifo mess with something simpler.

lnussel:
I was thinking of sharing modules between jeos-firstboot and sdbootutil too. Needs little duplication like d(), the file with the actual dialogs can then be sourced by different entry points.

files/usr/share/jeos-firstboot/modules/ssh_enroll
```diff
@@ -0,0 +1,131 @@
...
+
+	# Start the server in the background and show a dialog with information while it's running
+	# and offer the option to cancel the process. Unfortunately bash doesn't have a way to wait for
+	# one of multiple processes to exit and get information about which process exited and how.
+	# To work around that, use subprocesses with FIFOs: One to signal once they're up and can handle
+	# SIGTERM properly and one to signal that it exited and how.
+	# See also the welcome-screen implementation.
+	mkfifo "${tmpdir}/serverfifo"
+	mkfifo "${tmpdir}/dialogfifo"
+	mkfifo "${tmpdir}/exitfifo"
```
mgerstner, line 66: Actually I believe this *is* possible in bash like this:

```
sleep $(( $RANDOM % 20 ))  &
SLEEP1_PID=$!
sleep $(( $RANDOM % 20 )) &
SLEEP2_PID=$!

wait -n -p which %1 %2
res=$?

if [ "$which" -eq "$SLEEP1_PID" ]; then
        echo "sleep1 ($which) exited with $res"
else
        echo "sleep2 ($which) exited with $res"
fi

wait
```

Without the FIFOs and sub shells the code will become simpler.

files/usr/share/jeos-firstboot/modules/ssh_enroll
```diff
@@ -0,0 +1,131 @@
...
+		pid=$!; trap 'kill $pid' SIGTERM; echo server >"${tmpdir}/serverfifo"; wait $pid; echo server $? >>"${tmpdir}/exitfifo"
+	) &
+	serverpid=$!
+
+	{
+		echo $"Please connect to one of the addresses with ssh:"
+		hostname -I
+		echo
+		echo $"Please verify that the host key matches:"
+		for i in /etc/ssh/ssh_host_*key.pub; do ssh-keygen -l -f "$i"; done | awk '{ print $2" "$4 }'
```
mgerstner, line 79: Maybe you can use `ssh-keygen -lv` there in some way to also display the **visual host key** ASCIII art. I find this a lot easier to compare. You can also add a hint how to invoke `ssh` to display the visual host key upon connect (`ssh -o "VisualHostKey yes"`).

files/usr/share/jeos-firstboot/modules/ssh_enroll
```diff
@@ -0,0 +1,131 @@
...
+		return 0
+	fi
+
+	items=()
+	lineno=1
+	state=on
+	while read -r type key comment; do
+		items+=("$lineno" "${type} ${key:0:8}...${key: -20:20}" "$state")
+		lineno=$((lineno+1))
+		state=off # Only import the first key by default
```
mgerstner, line 19: Why import the first key by default? Is there some kind of priority ordering in SSH clients that makes the first key more fitting than the later ones?

I would not set any default here. Credentials are too sensitive to make any conclusions. The user should explicitly select the ones he wants.

files/usr/share/jeos-firstboot/modules/ssh_enroll
Vogtinator: Nice, I only had a look at the POSIX/sh wait documentation. This might be enough to work for this case.

In another branch (`ssh_enroll_simple`) I tried an even simpler method which requires the user to use Ctrl-C to abort the process, which is the one I'm currently favoring. I'll try this out in any case.

files/usr/share/jeos-firstboot/modules/ssh_enroll
Vogtinator: I thought about that as well but displaying the random art for all key types and also the ssh client command line in addition to the current dialog content is a bit much, especially if an 80x25 terminal is used.

files/usr/share/jeos-firstboot/modules/ssh_enroll
Vogtinator: > Why import the first key by default? Is there some kind of priority ordering in SSH clients that makes the first key more fitting than the later ones?

Yes, three reasons:

* It's the order defined in the ssh config file, i.e. by the user
* The first key is tried first, i.e. it's slightly quicker to log in
* If an SSH agent is used, the agent provided keys are first, i.e. the key the user most likely wants to use

files/usr/share/jeos-firstboot/modules/ssh_enroll
mgerstner: Maybe you can display it at least optionally, if the terminal is large enough. It would be a major usability feature IMHO. It could also be a UI element, like a button, that leaves TUI mode and displays the random art on the text console.

Vogtinator:
Displaying the randomart is a bit much but not as bad as I thought:
Screenshot_20240125_085618

files/usr/share/jeos-firstboot/modules/ssh_enroll
```diff
@@ -0,0 +1,131 @@
...
+	start_sshd_again=0
+	if systemctl -q is-active sshd.service; then
+		systemctl stop sshd.service
+		start_sshd_again=1
+	fi
+
+	# Make sure host keys exist
+	sshd-gen-keys-start &>/dev/null
+
+	tmpdir=$(umask 077; mktemp -d)
```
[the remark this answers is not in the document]
Vogtinator: Done

files/usr/share/jeos-firstboot/modules/ssh_enroll
Vogtinator: > Nice, I only had a look at the POSIX/sh wait documentation. This might be enough to work for this case.

Yep, done.

> In another branch (`ssh_enroll_simple`) I tried an even simpler method which requires the user to use Ctrl-C to abort the process, which is the one I'm currently favoring. I'll try this out in any case.

The inclusion of randomart means that scrolling might be necessary, so the simpler method no longer works with that.

files/usr/share/jeos-firstboot/modules/ssh_enroll
Vogtinator: Done, see the PR comment

files/usr/share/jeos-firstboot/modules/ssh_enroll
Vogtinator: Ok, first key is no longer selected by default

Vogtinator:
Next I'll make it more clear that the message on the client is important. As @lnussel requested, I'll also try to make this a standalone `ssh-pairing` script and move it into https://github.com/Vogtinator/ssh-pairing and integrate it into jeos-firstboot that way. Should only require minimal code changes.

Vogtinator:
> Next I'll make it more clear that the message on the client is important. As @lnussel requested, I'll also try to make this a standalone ssh-pairing script and move it into https://github.com/Vogtinator/ssh-pairing and integrate it into jeos-firstboot that way. Should only require minimal code changes.

Done! Superseded by #116

PR #124, opened 2024-05-24
Use UTC as default TZ if there was no locale selection dialog (boo#1224212)

Vogtinator:
If there is only a single locale available on the system, the locale selection is skipped. In that case the timezone should default to UTC though, to avoid that for the common case of en_US it ends up as America/New_York.

This was the previous default by accident because live-langset-data used US/Eastern as TZ for en_US, but the timezone dialog does not handle aliases.


Vogtinator:
Works with automatic selection and manual selection of en_US and de_DE as expected

PR #132, opened 2025-07-02
Use xterm on WSL based hosts boo#1237756

lkocman:
* Fixes issue with invalid encoding in WSL/Windows 10 cmd.exe
* We don't do systemd trigger, therefore the system ENV can stick with TERM=linux

image

After the change

image

### DIFF
--- a/files/usr/share/jeos-firstboot/welcome-screen
+++ b/files/usr/share/jeos-firstboot/welcome-screen
@@ -48,8 +48,13 @@ on_all_consoles() {
 	# Needed to tell apart errors and escape
 	export DIALOG_ERROR=254
 	# The linux fbcon uses this, most serial consoles should be fine too
-	export TERM=linux
-
+	# Use xterm on WSL otherwise Windows 10 cmd.exe
+	# will run into issues with encoding in both text and borders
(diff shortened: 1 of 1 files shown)


lkocman:
Alternatively we could simply go with TERM=xterm ...

files/usr/share/jeos-firstboot/welcome-screen
```diff
@@ -48,8 +48,11 @@ on_all_consoles() {
 	# Needed to tell apart errors and escape
 	export DIALOG_ERROR=254
 	# The linux fbcon uses this, most serial consoles should be fine too
-	export TERM=linux
-
+	TERM=linux
```
Vogtinator:
```
	# Use xterm on WSL otherwise Windows 10 cmd.exe
	# will run into issues with encoding in both text and borders
	if grep -qiw 'microsoft' /proc/version 2>/dev/null; then
		export TERM=xterm-256color
        else
		export TERM=linux
	fi
```

lkocman: applied. Thx!

Vogtinator (approved):

PR #135, opened 2026-03-21
Fix welcome_screen_with_console_switch undefined in jeos-config

### LINKED ISSUE #131: welcome_screen_with_console_switch is not defined when jeos-config is called
Some modules do an unconditional call to `welcome_screen_with_console_switch` in the `*_systemd_firstboot` entry point.  This function is defined normally, except when `jeos-config` is called.

Maybe an empty function should be defined early, and later redefined in `wellcome-screen`

anujagrawal380:
- Add a no-op stub for `welcome_screen_with_console_switch` in `jeos-firstboot-functions` so it is always defined, even when invoked via `jeos-config`
- The real implementation in `welcome-screen` overrides the stub when sourced by `jeos-firstboot`
- Prevents "command not found" errors when modules (e.g. NetworkManager) call `welcome_screen_with_console_switch` unconditionally in their `systemd_firstboot` hooks

Fixes #131

### DIFF
--- a/files/usr/share/jeos-firstboot/jeos-firstboot-functions
+++ b/files/usr/share/jeos-firstboot/jeos-firstboot-functions
@@ -36,6 +36,11 @@ list=
 password=''
 modules=()
 
+# Stub for welcome_screen_with_console_switch, redefined by the welcome-screen
+# file when sourced by jeos-firstboot. Modules may call this function
+# unconditionally, so it needs to be defined even when called via jeos-config.
+welcome_screen_with_console_switch() { :; }
+
 _find_modules() {
 	local f
 	for f in /etc/jeos-firstboot/modules/* "${jeos_prefix?}/share/jeos-firstboot/modules"/*; do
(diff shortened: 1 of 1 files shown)


anujagrawal380:
PTAL @Vogtinator

files/usr/share/jeos-firstboot/jeos-firstboot-functions
```diff
@@ -36,6 +36,11 @@ list=
 password=''
 modules=()
 
+# Stub for welcome_screen_with_console_switch, redefined by the welcome-screen
+# file when sourced by jeos-firstboot. Modules may call this function
+# unconditionally, so it needs to be defined even when called via jeos-config.
+welcome_screen_with_console_switch() { :; }
```
Vogtinator: Please move the definition to the end of the file

anujagrawal380: Done @Vogtinator

Vogtinator (approved):

PR #136, opened 2026-03-21
Add ConditionFirstBoot as alternative trigger for firstboot services

### LINKED ISSUE #98: Evaluate ConditionFirstBoot=yes
Currently `jeos-firstboot.service` runs iff `/var/lib/YaST2/reconfig_system` exists, which means that the file has to be created first.

For many image builds, systemd's built-in `ConditionFirstBoot` is sufficient and we can make use of that as well.

To still allow reconfiguration of the system and drop-in compatibility with `YaST2-Firstboot.service` we can try to use `ConditionFirstBoot=|yes` alongside `ConditionPathExists=|/var/lib/YaST2/reconfig_system`.

anujagrawal380:
- Add `ConditionFirstBoot=|yes` alongside `ConditionPathExists=|/var/lib/YaST2/reconfig_system` to both `jeos-firstboot.service` and `jeos-firstboot-snapshot.service`
- The `|` prefix makes them OR conditions, so the service triggers if either condition is met
- This removes the need to create `/var/lib/YaST2/reconfig_system` for image builds where systemd's built-in first boot detection is sufficient
- Backward compatibility with YaST2 and manual reconfiguration is preserved

### Ignition/Combustion compatibility

To prevent the firstboot wizard from running on systems already configured by ignition or combustion, runtime guards are added in `/usr/sbin/jeos-firstboot`:

- **Ignition**: Checks `/etc/.ignition-result.json` for `userConfigProvided: true` (mirrors the logic in `ignition-remove-reconfig_system.service`)
- **Combustion**: Checks for `/var/lib/combustion/done` marker file (requires a corresponding change in `openSUSE/combustion` to write this marker when running with `--complete`)
- The combustion marker is cleaned up by `jeos-firstboot-snapshot.service` alongside the existing `/var/lib/YaST2/reconfig_system` removal

Fixes #98

### DIFF
--- a/files/usr/lib/systemd/system/jeos-firstboot-snapshot.service
+++ b/files/usr/lib/systemd/system/jeos-firstboot-snapshot.service
@@ -9,7 +9,8 @@ After=apparmor.service local-fs.target plymouth-start.service YaST2-Second-Stage
 Conflicts=plymouth-start.service
 Before=getty@tty1.service serial-getty@ttyS0.service serial-getty@ttyS1.service serial-getty@ttyS2.service
 Before=display-manager.service
-ConditionPathExists=/var/lib/YaST2/reconfig_system
+ConditionPathExists=|/var/lib/YaST2/reconfig_system
+ConditionFirstBoot=|yes
 # The configuration is already done - so this doesn't make much sense
 #OnFailure=poweroff.target
 
--- a/files/usr/lib/systemd/system/jeos-firstboot-snapshot.service
+++ b/files/usr/lib/systemd/system/jeos-firstboot-snapshot.service
@@ -25,6 +25,7 @@ Type=oneshot
 RemainAfterExit=yes
 # In Pre - if creation fails, don't do the configuration again
 ExecStartPre=/usr/bin/rm -f /var/lib/YaST2/reconfig_system
+ExecStartPre=/usr/bin/rm -rf /var/lib/combustion/done
 ExecStart=/usr/sbin/jeos-firstboot-snapshot
 
 [Install]
--- a/files/usr/lib/systemd/system/jeos-firstboot-snapshot.service
+++ b/files/usr/lib/systemd/system/jeos-firstboot-snapshot.service
@@ -25,7 +25,6 @@ Type=oneshot
 RemainAfterExit=yes
 # In Pre - if creation fails, don't do the configuration again
 ExecStartPre=/usr/bin/rm -f /var/lib/YaST2/reconfig_system
-ExecStartPre=/usr/bin/rm -rf /var/lib/combustion/done
 ExecStart=/usr/sbin/jeos-firstboot-snapshot
 
 [Install]
--- a/files/usr/lib/systemd/system/jeos-firstboot.service
+++ b/files/usr/lib/systemd/system/jeos-firstboot.service
@@ -9,7 +9,8 @@ After=apparmor.service local-fs.target plymouth-start.service YaST2-Second-Stage
 Conflicts=plymouth-start.service
 Before=getty@tty1.service serial-getty@hvc0.service serial-getty@ttyS0.service serial-getty@ttyS1.service serial-getty@ttyS2.service serial-getty@ttyAMA0.service
 Before=display-manager.service
-ConditionPathExists=/var/lib/YaST2/reconfig_system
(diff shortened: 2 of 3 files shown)


anujagrawal380:
PTAL @Vogtinator

Vogtinator:
Unfortunately this will break setups which use ignition and/or combustion for unattended configuration: Those delete `/var/lib/YaST2/reconfig_system`: https://build.opensuse.org/projects/openSUSE:Factory/packages/ignition/files/ignition-remove-reconfig_system.service?expand=1 https://github.com/openSUSE/combustion/blob/c60eaacde224b30fac69bbf4c81b526a33281445/combustion#L336

We'd have to change the mechanism there.

For ignition, a check that looks at `/etc/.ignition-result.json` could be added to `/usr/sbin/jeos-firstboot`, but for combustion I'm not sure.

anujagrawal380:
> Unfortunately this will break setups which use ignition and/or combustion for unattended configuration: Those delete `/var/lib/YaST2/reconfig_system`: https://build.opensuse.org/projects/openSUSE:Factory/packages/ignition/files/ignition-remove-reconfig_system.service?expand=1 https://github.com/openSUSE/combustion/blob/c60eaacde224b30fac69bbf4c81b526a33281445/combustion#L336
> 
> We'd have to change the mechanism there.
> 
> For ignition, a check that looks at `/etc/.ignition-result.json` could be added to `/usr/sbin/jeos-firstboot`, but for combustion I'm not sure.

Thanks for the feedback @Vogtinator. You're right that this would break ignition/combustion setups since they suppress jeos-firstboot by deleting `/var/lib/YaST2/reconfig_system`, which isn't sufficient when `ConditionFirstBoot=|yes` also triggers the service.
I've added runtime guards in `/usr/sbin/jeos-firstboot` (following the existing cloud-init pattern) to skip execution when:
- Ignition already configured the system -- checks `/etc/.ignition-result.json` for `"userConfigProvided": true`, mirroring the logic in `ignition-remove-reconfig_system.service`
- Combustion already configured the system -- checks for a `/var/lib/combustion/done` marker file
The combustion marker is cleaned up by `jeos-firstboot-snapshot.service` alongside the existing `reconfig_system` removal.

This does require a small corresponding change in https://github.com/openSUSE/combustion to write the marker when running with --complete:
```                                                                 
  if [ "${complete}" -eq 1 ]; then                                
      rm -f /sysroot/var/lib/YaST2/reconfig_system                                                                                                                                                                   
      mkdir -p /sysroot/var/lib/combustion && touch /sysroot/var/lib/combustion/done                                                                                                                                 
  fi
```
I can open that PR against combustion as well if this approach looks good to you.

files/usr/lib/systemd/system/jeos-firstboot-snapshot.service
```diff
@@ -24,6 +25,7 @@ Type=oneshot
 RemainAfterExit=yes
 # In Pre - if creation fails, don't do the configuration again
 ExecStartPre=/usr/bin/rm -f /var/lib/YaST2/reconfig_system
+ExecStartPre=/usr/bin/rm -rf /var/lib/combustion/done
```
Vogtinator: Doesn't really make sense

Vogtinator:
> I can open that PR against combustion as well if this approach looks good to you.

Something like that looks good, but for consistency it should maybe be `echo '{}' > /sysroot/etc/.combustion-result.json`.

I'll mark the PR as a draft as it needs corresponding changes in combustion.

It might even have to wait until the combustion changes made it into relevant distributions...

anujagrawal380:
> > I can open that PR against combustion as well if this approach looks good to you.
> 
> Something like that looks good, but for consistency it should maybe be `echo '{}' > /sysroot/etc/.combustion-result.json`.
> 
> I'll mark the PR as a draft as it needs corresponding changes in combustion.
> 
> It might even have to wait until the combustion changes made it into relevant distributions...

Updated, switched to `/etc/.combustion-result.json` for consistency and removed the ExecStartPre cleanup.
                                                                                                                                                                                                                     
I've opened an issue on openSUSE/combustion for the corresponding change: https://github.com/openSUSE/combustion/issues/46
Will open a PR there as well. How should we ideally rollout after combustion changes merge?

